PR #7, opened 2021-03-11
ERC721

ItsNickBarry:


sunnya97:
The `transferFrom`, `safeTransferFrom`, and `approve` functions should all be payable.

OpenZeppelin had this incorrect, and an issue was opened for it here:  https://github.com/OpenZeppelin/openzeppelin-contracts/issues/2610

I also submitted a PR fix for it here:  https://github.com/OpenZeppelin/openzeppelin-contracts/pull/2618

Amxx:
Please don't commit any more work to this PR until further notice.

I understand you frustration with the function not being payable, but what you are proposing would cause tremendous breaking changes. The use of a storage layout, in particular, would make this version not "storage-compatible" with existing instances. This would prevent user upgrading their contracts.

ItsNickBarry:
Thanks for the information.  Seems like using nonpayable functions is technically compliant:
> The above interfaces include explicit mutability guarantees for each function. Mutability guarantees are, in order weak to strong: payable, implicit nonpayable, view, and pure. Your implementation MUST meet the mutability guarantee in this interface and you MAY meet a stronger guarantee. For example, a payable function in this interface may be implemented as nonpayble (no state mutability specified) in your contract.

However, since SolidState is intended to provide fundamental building blocks, it makes sense to be as flexible as possible.  I don't really want to revert the transaction as in your OpenZeppelin PR, because the functions in question are sensitive, and I don't want users to copy and paste (or worse) when overriding.

Here are some alternatives:
* Add an unimplemented `withdrawEther` function to the `ERC721` contract for the user to implement,  but omit it from the `ERC721Base` contract; the latter is intended for more advanced users.
* Add an `_onPaymentReceived` hook that can be used like the `_beforeTokenTransfer` hook.
* Do both of these.

Not sure what option is best yet.

Dominus-Esports (commented):
Looks good to me

PR #12, opened 2021-04-21
add placeholder tests

zimmah:


test/token/ERC1155/ERC1155Enumerable.behavior.js
```diff
@@ -4,14 +4,18 @@ const describeBehaviorOfERC1155Base = require('./ERC1155Base.behavior.js');
...
   describe('::ERC1155Enumerable', function () {
     let instance;
 
     beforeEach(async function () {
-      instance = await ethers.getContractAt('ERC20Metadata', (await deploy()).address);
+      instance = await ethers.getContractAt('ERC1155EnumerableMock', (await deploy()).address);
+      const [signer] = await ethers.getSigners();
+      const id = ethers.constants.Zero;
+      const amount = ethers.constants.Two;
+      instance.mint(signer.address, id, amount, ethers.utils.randomBytes(0));
```
ItsNickBarry, line 18: Missing `await`.

contracts/token/ERC1155/ERC1155EnumerableMock.sol
```diff
@@ -11,4 +11,13 @@ contract ERC1155EnumerableMock is ERC1155Enumerable {
     ERC165Storage.layout().setSupportedInterface(type(IERC165).interfaceId, true);
     ERC165Storage.layout().setSupportedInterface(type(IERC1155).interfaceId, true);
   }
+
+  function mint (
+  address account,
```
ItsNickBarry, line 16: Indentation

test/token/ERC1155/ERC1155Enumerable.behavior.js
```diff
@@ -20,22 +24,36 @@ const describeBehaviorOfERC20Metadata = function ({ deploy }, skips) {
     }, skips);
 
     describe('#totalSupply', function () {
-      it('returns supply of given token');
+      it('returns supply of given token', async function(){
+        let id = ethers.constants.Zero;
+        expect(await instance.callStatic.totalSupply(id)).to.equal(2);
```
ItsNickBarry, line 29: Don't hardcode the value of `2` if it's defined in a constant above (`amount`, line 17).

ItsNickBarry (changes requested):
@zimmah Instead of minting the token in the `beforeEach` block, mint within each of the test functions.  Then run the same test before and after the minting, so that the change in state is tested as well as the final state.

ItsNickBarry (approved):

PR #13, opened 2021-04-24
add erc1155Base transfer tests

zimmah:
missing erc20base tests


ItsNickBarry (changes requested):
@zimmah The `mint` and `burn` functions need to be passed to the behavior tests everywhere they're called.  For example, here: https://github.com/zimmah/solidstate-solidity/blob/799926c2af0b9847dfa4d90771bce28db2357bb6/test/token/ERC1155/ERC1155Enumerable.behavior.js#L19.

PR #18, opened 2021-05-22
Added error messages on ERC1155 and ERC20

Lumyo:
- Added insufficient balance message on ERC1155 
- Added exceeds allowance error message on ERC20

Adds only ~60 gas usage on the function calls, and allows to check for error message when writing tests, to ensure reason of revert


ItsNickBarry:
Missing updates to the error messages in the `__internal` tests in `ERC1155Base.js`.  Also if we're checking for overflows explicitly, we can probably use `unchecked` blocks in those functions to save gas.

Lumyo:
@ItsNickBarry Pushed both fixes

PR #19, opened 2021-05-25
Typescript upgrade

Lumyo:
Upgrade the repository to use typescript + typechain for fully typed tests


contracts/signature/ERC1271Stored.sol
```diff
@@ -11,7 +11,7 @@ import './ERC1271StoredStorage.sol';
 abstract contract ERC1271Stored is ERC1271Base {
   function _isValidSignature (
     bytes32 hash,
-    bytes memory
+    bytes memory signature
```
ItsNickBarry, line 14: Why name the variable?  It's unused and therefore will cause a compilation warning.

contracts/signature/ERC1271StoredMock.sol
```diff
@@ -8,7 +8,7 @@ import './ERC1271Stored.sol';
 contract ERC1271StoredMock is ERC1271Stored {
   constructor (
     bytes32 hash,
-    bytes memory
+    bytes memory signature
```
ItsNickBarry, line 11: Why name the variable? It's unused and therefore will cause a compilation warning.

copyAbi.ts
```diff
@@ -0,0 +1,237 @@
+import fs from 'fs';
```
ItsNickBarry, line 1: What is the purpose of this?  Can it be accomplished using [hardhat-abi-exporter](https://github.com/ItsNickBarry/hardhat-abi-exporter)?

ItsNickBarry: If not, let's move this to a post-compilation hook.  I don't want to have to modify this list every time a contract is added or removed.

hardhat.config.ts
```diff
@@ -43,9 +44,9 @@ module.exports = {
   networks: {
     generic: {
       // set URL for external network
-      url: `${ process.env.URL }`,
+      url: `${process.env.URL}`,
```
ItsNickBarry, line 47: Going to need padding inside of these brackets.  Let's set a prettier or eslint rule at some point.

package.json
```diff
@@ -3,19 +3,33 @@
   "private": true,
   "author": "Nick Barry",
   "license": "MIT",
+  "scripts": {
+    "compile": "hardhat compile && yarn copyAbi && yarn typechain",
```
ItsNickBarry, line 7: Prefer to override the compile task so that the native `hh compile` functions as expected.

package.json
```diff
@@ -3,19 +3,33 @@
...
     "@nomiclabs/hardhat-ethers": "^2.0.2",
     "@nomiclabs/hardhat-waffle": "^2.0.1",
     "@solidstate/spec": "link:spec",
+    "@typechain/ethers-v5": "^7.0.0",
+    "@types/chai": "^4.2.18",
+    "@types/mocha": "^8.2.2",
+    "@types/node": "^15.6.0",
+    "@types/rimraf": "^3.0.0",
     "chai": "^4.3.4",
-    "eslint": "^7.24.0",
```
ItsNickBarry, line 11: Why remove eslint?

package.json
```diff
@@ -3,19 +3,33 @@
...
-    "eslint-plugin-mocha": "^8.1.0",
```
ItsNickBarry, line 12: If we're removing this, need to remove many instances of `// eslint-disable-next-line`.

copyAbi.ts
ItsNickBarry: Note that ABIs are already exported to `abi` directory and published as `@solidstate/abi`.

test/factory/CloneFactory.ts
```diff
@@ -23,10 +25,8 @@ describe('CloneFactory', function () {
 
           await instance['deployClone()']();
 
-          expect(
-            await ethers.provider.getCode(address)
-          ).to.equal(
-            await ethers.provider.getCode(instance.address)
+          expect(await ethers.provider.getCode(address)).to.equal(
```
ItsNickBarry, line 28: This format is painful to read.  The work of `prettier`, I assume?

spec/token/ERC20/ERC20Base.behavior.ts
```diff
@@ -0,0 +1,177 @@
...
+    let receiver: SignerWithAddress;
+    let sender: SignerWithAddress;
+    let instance: ERC20Base;
+
+    before(async function () {
+      [holder, spender, receiver, sender] = await ethers.getSigners();
+    });
+
+    beforeEach(async function () {
+      instance = await deploy();
```
ItsNickBarry, line 33: How does Typescript handle casting of contracts?  If a third party contract has both `balanceOf(address)` and a `balanceOf(address,uint256)` functions, for example, will the calls in this file fail due to the overload?

tsconfig.json
```diff
@@ -0,0 +1,11 @@
...
+    "compilerOptions": {
+        "target": "es2018",
+        "module": "commonjs",
+        "strict": true,
+        "esModuleInterop": true,
+        "outDir": "dist"
+    },
+    "include": ["./scripts", "./test"],
+    "files": ["./hardhat.config.ts"]
+}
```
ItsNickBarry, line 11: Many files lack newlines, and Github complains.

contracts/signature/ERC1271Stored.sol
Lumyo: Not having this variable named prevents typechain from generating the typed contracts.

image

contracts/signature/ERC1271StoredMock.sol
Lumyo: Not having this variable named prevents typechain from generating the typed contracts.

image

copyAbi.ts
Lumyo: It requires the full artifacts.
I just found out about this hardhat plugin : https://www.npmjs.com/package/hardhat-typechain
I'll replace with that

package.json
Lumyo: Will be taken care with the plugin https://www.npmjs.com/package/hardhat-typechain

hardhat.config.ts
Lumyo: Prettier doesnt have an option to add padding inside template litterals, only for brackets of objects

package.json
Lumyo: I feel like eslint is redundant when using typescript, as typescript compiler already detects code errors (+ type errors which eslint doesnt check) and prettier takes care of the code formatting.

test/factory/CloneFactory.ts
Lumyo: Yes, prettier.

spec/token/ERC20/ERC20Base.behavior.ts
Lumyo: Good point.
So by default, types with method signatures are only generated when there is an overload, but typechain has a `--always-generate-overloads` options, which generate types with method signature for everything.

So I'll update to use this flag, and will use everywhere the function call through the method signature, which will solve this.

contracts/signature/ERC1271Stored.sol
ItsNickBarry: Do you know of a fundamental reason for this, or should I open an issue in the TypeChain repository?

ItsNickBarry: For now I think the solution is to add a line `signature = signature`.  I don't think this costs gas if the optimizer is turned on, though not sure.

test/factory/CloneFactory.ts
ItsNickBarry: Has prettier formatting been applied to all files?  Some are worse than others.

ItsNickBarry:
@Lumyo How will this affect publication to NPM?

package.json
ItsNickBarry: I think you'd be right, if prettier were actually capable of making things prettier.  For now the removal of eslint is fine, as it reduces complexity, but I might add it again at some point.

contracts/signature/ERC1271Stored.sol
Lumyo: No, not aware of a particular reason for this

test/factory/CloneFactory.ts
Lumyo: Yes it has

ItsNickBarry:
Found a workaround to the failing Factory tests.  Underlying issue reported here: https://github.com/nomiclabs/hardhat/issues/1488

Lumyo:
@ItsNickBarry 
I think this is the cleanest way to handle the TS build : https://github.com/solidstate-network/solidstate-solidity/pull/19/commits/e7d5978cefcc5e34f1c8c8b580734725f11498ac

- Moved `typechain` generation to `spec/typechain`
- Created a typescript path mapping from `@solidstate/typechain` to `spec/typechain`

Everywhere, we import typechain classes through `@solidstate/typechain` but this is not a separate package, so no need to publish it / link it for local development. The typescript compiler will remap the paths on compilation, and they will directly be included in `@solidstate/spec` which is the only published package which needs them.

If later we want to deploy a separate `@solidstate/typechain` package, it will be easy as the imports are done like that everywhere already (Would just need to remove the TS path mapping)

Lumyo:
Actually, despite what i was thinking,, the path alias mentioned above are not remapped on compilation, so I'm gonna use relative import in `spec`

ItsNickBarry:
Added a types package here: https://github.com/solidstate-network/solidstate-solidity/pull/22

The compilation override to copy the `typechain` directory isn't clean, but I had to do the same thing for the ABI package anyway.

Lumyo:
@ItsNickBarry Can we merge this ?

ItsNickBarry:
If all internal packages are set up to be published on NPM and imported by both JS and TS, then yes.  Can you confirm?  I was going to test by publishing to a local package registry like verdaccio.

Lumyo:
`spec` and `library` packages typescript configs should be fine.
Just compile both using `tsc` and publish including the `dist` folder, which contain JS code + types (so will work with both JS and TS)
Not sure if there is anything else specific to NPM which needs to be done

PR #21, opened 2021-05-27
use typechain hardhat plugin

ItsNickBarry:
@Lumyo This branch includes a mostly-working installation of the Typechain Hardhat plugin.  However, after running `hh clean`, `hh test` fails the first time unless `hh compile` has been run explicitly.  It works as expected on subsequent calls.  Is this the issue that you mentioned?


.gitignore
```diff
@@ -11,6 +11,4 @@ yarn.lock
 
 .env
 
-dist
```
Lumyo, line 14: You need to leave `dist` in the `.gitignore`, as this folder is generated by the typescript compiler and should not be commited

Lumyo:
I had some other error during compilation, but I dont have it anymore. 
So it was probably related to something else

.gitignore
ItsNickBarry: That's what I thought, but I don't see `dist` anywhere.  When is it generated?

Lumyo: At the root of the project when you run the typescript compiler through `tsc`

ItsNickBarry: When is `tsc` run?

Lumyo: You can run it anytime when you wanna check that your code compiles without error (some errors might not be displayed by the IDE, but will be caught by the TS compiler) or when you wanna transpile your code to JS for deployment / publishing

ItsNickBarry: Okay, fixed.

PR #22, opened 2021-05-28
add types package

ItsNickBarry:


ItsNickBarry:
This includes the factory classes as well as the types, but I don't think we need to worry about that right now.

Is there anything we need to add to `package.json` to allow other TS projects to import the types?

Lumyo:
@ItsNickBarry  This solution is not ideal because generated files should not be committed to the repo.
Typechain even mentions it in their FAQ
image

Also, when types are published separately from the package they are used in, they should be published through https://github.com/DefinitelyTyped/DefinitelyTyped

ItsNickBarry:
So the generated files should be committed to a different repository?  That's a bit contradictory, and significantly more work.

That FAQ entry is true for top-level contract repositories like `premia-contracts`, but I don't think it applies here.  In some cases, the contracts referenced by `spec` might not be available locally, so typechain will have no way of generating the types.  Publishing types to DefinitelyTyped would not fit in well with the lerna process.

However, I may have misunderstood how the `tsc` build process works.  It looks like both `.ts` and `.js` entry points are generated, and all relative imports will be included in the `dist` directories.  So it might not be necessary to publish a types package at all.

zimmah:
I think generated files don't need to be in any repo, because they should be generated locally as part of the install / post-install.

PR #29, opened 2021-06-09
Erc721tests

zimmah:


ItsNickBarry:
@zimmah I'm not sure what happened with the commit history here.  Branches are about even, so why so many commits?  Should I look at #28 instead?  That one needs an updated target branch if so.

ItsNickBarry:
Looks good at first glance, merging into WIP ERC721 branch.

PR #36, opened 2021-07-10
Run Prettier on commit

ItsNickBarry:


ItsNickBarry:
Before merge, format entire project.

PR #37, opened 2021-07-10
Hacken audit response

ItsNickBarry:
Points raised in the Hacken audit:
* [ ] Not enough comments
* [x] Return value ignored in `ECDSAMultisigWalletStorage`
* [x] Return value ignored in `ERC1155Enumerable`
* [x] Code layout (?)
* [x] Some contracts should be declared `abstract`
* [x] Constant naming conventions

### DIFF
--- a/abi/Diamond.json
+++ b/abi/Diamond.json
@@ -1,10 +1,5 @@
 [
   {
-    "inputs": [],
-    "stateMutability": "nonpayable",
-    "type": "constructor"
-  },
-  {
     "anonymous": false,
     "inputs": [
       {
--- a/abi/ERC1404.json
+++ b/abi/ERC1404.json
@@ -292,7 +292,7 @@
     "inputs": [
       {
         "internalType": "address",
-        "name": "sender",
+        "name": "holder",
         "type": "address"
       },
       {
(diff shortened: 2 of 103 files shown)


ItsNickBarry:
Regarding the `abstract` keyword:

The reasoning behind avoiding its use for certain contracts, such as `Ownable`, is that they may be deployed standalone and their external functions referenced by a diamond proxy.  In fact, a few contracts which are currently declared `abstract` can potentially be used in this way:
* `ERC165`
* `ManagedProxyOwnable`
* most `token` contracts

ItsNickBarry:
Regarding the unchecked `EnumerableSet` return value in `ERC1155Enumerable`:

This will only ever be an issue if a developer accesses the storage layout directly and corrupts the data.  In such a situation, a `revert` could potentially be worse than a silent failure, as a contract could theoretically become locked.  Perhaps a `revert` would be more visible in testing?  It's difficult to determine how valuable this would be.  Instead, the solution might be to standardize storage layout access globally using getters and setters where appropriate, and renaming storage variables to something less accessible (`balances` -> `_balances`) to indicate that direct access is not recommended.

ItsNickBarry:
Inline documentation is an ongoing process.  Improvements have been made as a part of this PR, but comprehensive coverage is considered to be out of scope due to time constraints.

PR #45, opened 2021-08-26
Visibility layers

### LINKED ISSUE #42: Define `internal` functions in separate contracts
Interaction between diamond facets is best done using `internal` functions, exposed either through inheritance or shared libraries.  Most contracts in this repository define both `internal` and `external` functions, leading to duplicate unused `external` functions across facets.  Some `internal` functions, such as ERC20 `_mint`, cannot be defined in libraries because they emit events which are defined in interfaces, from which libraries cannot inherit.

Therefore, it might make sense to divide each contract into separate parts, organized by visibility level:
* storage library
* interface
* internal functions
* external functions

ItsNickBarry:
closes #42


ItsNickBarry:
@Lumyo This is almost ready to merge; changes have been applied to everything in the `token` directory.  See description added to the `contracts` package README for context: https://github.com/solidstate-network/solidstate-solidity/blob/05a65559a147d1ab3ff0b9555942cf589fded3d3/contracts/README.md

No need to review each file individually, but I'd like your opinion about whether this architecture change makes sense and whether it could be improved with additional ideas.

contracts/token/ERC1155/base/ERC1155BaseInternal.sol
```diff
@@ -2,102 +2,31 @@
...
+   * @param account address to query
+   * @param id token to query
+   * @return token balance
    */
-  function balanceOf (
+  function _balanceOf (
     address account,
     uint id
-  ) virtual override public view returns (uint) {
-    require(account != address(0), 'ERC1155: balance query for the zero address');
```
Lumyo, line 24: I probably would keep the require in the internal function, as querying the balance of the 0 address is probably a behavior that we would never want even in the internal function, and check is cheap anyway
(That particular error already helped us debugging the queue before)

contracts/token/ERC721/base/ERC721Base.sol
```diff
@@ -0,0 +1,94 @@
...
+  using EnumerableSet for EnumerableSet.UintSet;
+
+  /**
+   * @inheritdoc IERC721
+   */
+  function balanceOf (address account) override public view returns (uint) {
+    require(account != address(0), 'ERC721: balance query for the zero address');
+    return ERC721BaseStorage.layout().holderTokens[account].length();
+  }
+
```
Lumyo, line 28: Split with internal `_balanceOf` function, for consistency with ERC1155 ?

Lumyo (approved):

contracts/token/ERC721/base/ERC721Base.sol
ItsNickBarry: I think yes, but where do we stop?  Should every external function have a corresponding internal function?

PR #53, opened 2021-11-01
Improve ERC721 Integrations

ItsNickBarry:
3 functions modified to be chained through:
* external contract
* internal contract
* storage library

~19k additional deployment gas
~100 additional function call gas

### DIFF
--- a/contracts/token/ERC721/ERC721.sol
+++ b/contracts/token/ERC721/ERC721.sol
@@ -21,11 +21,12 @@ abstract contract ERC721 is
      * @inheritdoc ERC721BaseInternal
      */
     function _handleApproveMessageValue(
-        address,
-        uint256,
+        address operator,
+        uint256 tokenId,
         uint256 value
     ) internal virtual override {
         require(value == 0, 'ERC721: payable approve calls not supported');
+        super._handleApproveMessageValue(operator, tokenId, value);
     }
 
     /**
@@ -33,12 +34,13 @@ abstract contract ERC721 is
      * @inheritdoc ERC721BaseInternal
      */
     function _handleTransferMessageValue(
-        address,
-        address,
-        uint256,
+        address from,
+        address to,
+        uint256 tokenId,
         uint256 value
     ) internal virtual override {
         require(value == 0, 'ERC721: payable transfer calls not supported');
+        super._handleTransferMessageValue(from, to, tokenId, value);
     }
 
     /**
--- a/contracts/token/ERC721/base/ERC721BaseInternal.sol
+++ b/contracts/token/ERC721/base/ERC721BaseInternal.sol
@@ -184,7 +184,7 @@ abstract contract ERC721BaseInternal is IERC721Internal {
         address operator,
         uint256 tokenId,
         uint256 value
-    ) internal virtual;
+    ) internal virtual {}
 
     /**
      * @notice ERC721 hook, called before externally called transfers for processing of included message value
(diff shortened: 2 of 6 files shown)


contracts/token/ERC721/ERC721.sol
```diff
@@ -21,24 +21,26 @@ abstract contract ERC721 is
...
      */
     function _handleApproveMessageValue(
-        address,
-        uint256,
+        address operator,
+        uint256 tokenId,
         uint256 value
     ) internal virtual override {
         require(value == 0, 'ERC721: payable approve calls not supported');
+        super._handleApproveMessageValue(operator, tokenId, value);
```
NouDaimon, line 29: Why super.? It seems the implementation in ERC721BaseInternal is empty. 

Please explain.

contracts/token/ERC721/ERC721.sol
```diff
@@ -21,24 +21,26 @@ abstract contract ERC721 is
...
-        address,
-        address,
-        uint256,
+        address from,
+        address to,
+        uint256 tokenId,
         uint256 value
     ) internal virtual override {
         require(value == 0, 'ERC721: payable transfer calls not supported');
+        super._handleTransferMessageValue(from, to, tokenId, value);
```
NouDaimon, line 43: Same comment as above.

contracts/token/ERC721/base/ERC721BaseInternal.sol
```diff
@@ -53,25 +57,26 @@ abstract contract ERC721BaseInternal is IERC721Internal {
         returns (bool)
     {
         require(
-            ERC721BaseStorage.exists(tokenId),
+            ERC721BaseStorage.layout().exists(tokenId),
```
NouDaimon, line 60: For consistency perhaps use the same approach as above in  `_getApproved` and below in `_mint`?

NouDaimon (commented):
All changes seem appropriate; just asked for some clarification on 1-2 functions.

Generally, it is a good idea to refactor functionality into internals - the contracts themselves become more easily extensible & modular in that way.

contracts/token/ERC721/ERC721.sol
ItsNickBarry: If multiple contracts inherit from `ERC721Base` and override this function, they all must call `super` to ensure that all override code is run.

contracts/token/ERC721/base/ERC721BaseInternal.sol
ItsNickBarry: In what sense?  Defining the `l` variable on a separate line?  Generally I only do that when it needs to be accessed multiple times.

NouDaimon: Oh ofcourse. My oversight, thanks for clarifying.

contracts/token/ERC721/ERC721.sol
NouDaimon: But does ERC721Base need `super` in itself? The contract it is inheriting from has no implementation.

ItsNickBarry: Not sure what you mean.  `ERC721Base` does not call `super`.  In general, the idea is that someone can import `ERC721BaseInternal` as well as another contract that overrides the hook, and both must call `super` as a convention to ensure that all implementations are included, regardless of inheritance order.  It's not particularly useful here because the hook will be rarely used, but I chose this approach because it supports `payable` ERC721 functions (as specified in the standard) using the preexisting `_beforeTokenTransfer` hook pattern.

NouDaimon: Thank you for the explanation.

PR #59, opened 2021-12-03
`ERC1155EnumerableInternal` function logic

ItsNickBarry:


ItsNickBarry:
@Lumyo Merging and publishing, but worth a check.

PR #68, opened 2022-01-07
Upgradeable proxy

NouDaimon:


contracts/proxy/upgradeable/UpgradeableOwnableProxyBase.sol
```diff
@@ -0,0 +1,35 @@
...
+        return UpgradeableOwnableProxyStorage.layout().implementation;
+    }
+
+    function _getOwner() internal view returns (address) {
+        return OwnableStorage.layout().owner;
+    }
+
+    function _setOwner(address owner) internal {
+        OwnableStorage.layout().setOwner(owner);
+    }
```
ItsNickBarry, line 34: These functions are very Ownable-specific, and should not be defined here (maybe they should be defined in `OwnableInternal`, but that is a question for later).

contracts/proxy/upgradeable/UpgradeableOwnableProxyBase.sol
```diff
@@ -0,0 +1,35 @@
...
+import { Proxy } from '../Proxy.sol';
+import { OwnableInternal } from '../../access/OwnableInternal.sol';
+import { OwnableStorage } from '../../access/OwnableStorage.sol';
+import { UpgradeableOwnableProxyStorage } from './UpgradeableOwnableProxyStorage.sol';
+
+abstract contract UpgradeableOwnableProxyBase is OwnableInternal, Proxy {
+    using UpgradeableOwnableProxyStorage for UpgradeableOwnableProxyStorage.Layout;
+    using OwnableStorage for OwnableStorage.Layout;
+
+    receive() external payable {}
```
ItsNickBarry, line 14: This may not be desired in all use cases, so should be removed.

contracts/proxy/upgradeable/UpgradeableOwnableProxyBase.sol
```diff
@@ -0,0 +1,35 @@
...
+
+    constructor(address implementation) {
```
ItsNickBarry, line 16: Do not define constructor in any `Base` contract.

contracts/proxy/upgradeable/UpgradeableOwnableProxy.sol
```diff
@@ -0,0 +1,20 @@
...
+
+pragma solidity ^0.8.0;
+
+import { UpgradeableOwnableProxyBase } from './UpgradeableOwnableProxyBase.sol';
+import { Proxy } from '../Proxy.sol';
+import { OwnableInternal } from '../../access/OwnableInternal.sol';
+import { OwnableStorage } from '../../access/OwnableStorage.sol';
+import { UpgradeableOwnableProxyStorage } from './UpgradeableOwnableProxyStorage.sol';
+
+contract UpgradeableOwnableProxy is UpgradeableOwnableProxyBase {
```
ItsNickBarry, line 11: Rename to `UpgradeableProxyOwnable` for parallel structure with `ManagedProxyOwnable`, etc.

ItsNickBarry (changes requested):

ItsNickBarry:
@NouDaimon Let's remove the `Base` contract layer, and make all contracts `abstract`.

Files needed:
* `UpgradeableProxy.sol`
* `UpgradeableProxyStorage.sol`
* `UpgradeableProxyOwnable.sol`
* `UpgradeableProxyMock.sol`
* `UpgradeableProxyOwnableMock.sol`

`UpgradeableProxyOwnable` should inherit from `OwnableInternal`, but should not set owner at any point.  For this reason, it should be `abstract`.

test/proxy/upgradeable/UpgradeableProxy.ts
```diff
@@ -0,0 +1,58 @@
...
+    deploy: async () => instance as any,
+    implementationFunction: 'owner()',
+    implementationFunctionArgs: [],
+  });
+
+  describe('__internal', function () {
+    describe('#_getImplementation', function () {
+      it('returns implementation address', async function () {
+        expect(await instance.callStatic.__getImplementation()).to.be
+          .properAddress;
```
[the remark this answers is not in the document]
ItsNickBarry: No need, because the `#_setImplementation` test validates that the correct address is returned.  Testing that here would really just be testing that the mock contract sets the right address (but we don't really care about the mock contract).

ItsNickBarry (approved):

PR #74, opened 2022-01-29
Overload test separation

NouDaimon:
@ItsNickBarry I checked out from the `override-removal` branch, so there are duplicate commits.

### DIFF
--- a/test/cryptography/ECDSA.ts
+++ b/test/cryptography/ECDSA.ts
@@ -42,7 +42,7 @@ describe('ECDSA', function () {
       it('todo');
     });
 
-    describe('#toEthSignedMessageHash', function () {
+    describe('#toEthSignedMessageHash(bytes32)', function () {
       it('returns hash of signed message prefix and message', async function () {
         const hash = ethers.utils.keccak256(ethers.utils.toUtf8Bytes('test'));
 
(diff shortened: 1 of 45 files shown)


ItsNickBarry:
@NouDaimon See also this part of the task:

>  Non-overloaded functions should also include the full function signature in the test suite description.

For example, `describe('#balanceOf', ...)` should be `describe('#balanceOf(address)', ...)`, and `describe('#totalSupply', ...)` should be `describe('#totalSupply()', ...)`.

test/cryptography/ECDSA.ts
```diff
@@ -12,36 +12,34 @@ describe('ECDSA', function () {
...
-          const sig = await signData(signer, data);
-
-          expect(
-            await instance.callStatic['recover(bytes32,bytes)'](
-              ethers.utils.solidityKeccak256(
-                ['string', 'bytes32'],
-                ['\x19Ethereum Signed Message:\n32', hash],
-              ),
-              sig,
+    describe('#recover(bytes32, bytes)', () => {
```
ItsNickBarry, line 15: No space between parameter types
```suggestion
    describe('#recover(bytes32,bytes)', () => {
```

ItsNickBarry (changes requested):
A few still need to be changed.  For example `#nomineeOwner` in `SafeOwnable.behavior.ts`.  Also, some tests are now failing, but I can't tell why.

ItsNickBarry:
Issue with failing tests fixed.

ItsNickBarry (approved):

PR #77, opened 2022-02-13
misc: optimize gas in EnumerableSet and EnumerableMap

antoncoding:
## Updates
Main idea: use `unchecked` to avoid duplicate underflow check on `keyIndex - 1` 

## Gas report
according to the gas report, this saves an average of 100~ gas on involved methods. Here are some examples:

| Contract | Method | Avg. gas used before | Avg. gas used after | 
| -------- | -------- | -------- | -------- |
| ERC1155EnumerableMock | __burn | 53484 | 53060 |
| ERC1155EnumerableMock | safeTransferFrom | 136119 | 135589 |
| ERC1155Mock | __burn | 53517 | 53093 |
| ERC1155Mock | safeBatchTransferFrom | 139048 | 138518 |

### DIFF
--- a/contracts/utils/EnumerableSet.sol
+++ b/contracts/utils/EnumerableSet.sol
@@ -197,7 +197,6 @@ library EnumerableSet {
 
         if (valueIndex != 0) {
             unchecked {
-                // don't need to check underflow because valueIndex must > 0.
                 uint256 index = valueIndex - 1;
 
                 bytes32 last = set._values[set._values.length - 1];
--- a/contracts/utils/EnumerableSet.sol
+++ b/contracts/utils/EnumerableSet.sol
@@ -196,13 +196,16 @@ library EnumerableSet {
         uint256 valueIndex = set._indexes[value];
 
         if (valueIndex != 0) {
-            uint256 index = valueIndex - 1;
+            uint256 index;
+            unchecked {
+                index = valueIndex - 1;
+            }
             bytes32 last = set._values[set._values.length - 1];
 
             // move last value to now-vacant index
 
             set._values[index] = last;
-            set._indexes[last] = index + 1;
+            set._indexes[last] = valueIndex;
 
             // clear last index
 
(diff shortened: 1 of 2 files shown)


contracts/utils/EnumerableSet.sol
```diff
@@ -196,13 +196,16 @@ library EnumerableSet {
...
+            }
             bytes32 last = set._values[set._values.length - 1];
 
             // move last value to now-vacant index
 
             set._values[index] = last;
-            set._indexes[last] = index + 1;
+            set._indexes[last] = valueIndex;
 
             // clear last index
```
ItsNickBarry, line 211: Should be okay to wrap this entire block in `unchecked`, including the `set._values.length - 1`, right?

antoncoding: good idea, this will also makes `map._indexes[last._key] = index + 1;` a bit better

contracts/utils/EnumerableSet.sol
```diff
@@ -196,14 +196,16 @@ library EnumerableSet {
...
 
-            // move last value to now-vacant index
+                bytes32 last = set._values[set._values.length - 1];
 
-            set._values[index] = last;
-            set._indexes[last] = index + 1;
+                // move last value to now-vacant index
 
+                set._values[index] = last;
+                set._indexes[last] = index + 1;
```
ItsNickBarry, line 207: I don't agree with the readability argument from the Discord chat.  Better to use the value already stored in a variable.

```suggestion
                set._indexes[last] = valueIndex;
```

ItsNickBarry: And with that change, we probably don't need a reference to `index`, because it will only be used once.

antoncoding: I don't feel strongly about readability re `set._indexes[last] = valueIndex;` vs `set._indexes[last] = index + 1;`, happy to go either way. 

But overall I do like the idea of removing `index` completely, feel like that makes it easier to follow the logic, without having to memorize 2 index-liked variables.

NouDaimon (approved):

ItsNickBarry:
@antoncoding Thanks, merging.  Will release a new version soon.

PR #81, opened 2022-03-20
Fix ERC1404 storage location

ItsNickBarry:
The ERC1404 storage location and storage library name do not conform to standards set by `ERC20Base`, etc.

@NouDaimon Would you take a quick look at the other storage libraries and make sure I didn't miss any more?


antoncoding (approved):

ItsNickBarry:
Thanks @antoncoding.  The question that remains is whether to merge a breaking change of this nature now, or wait until an explicitly breaking release like `1.0.0`.  Probably best to just get it over with, since all `0.0.x` releases can be expected to be breaking.

PR #95, opened 2022-05-10
added AccessControl

iam-dev:
Test need to be done


iam-dev:
btw where can I change the @solidstate/spec?
I would like  to add describeBehaviorOfAccessControl

ItsNickBarry:
That package is located in the `spec` directory of this repository.

package.json
```diff
@@ -13,6 +13,7 @@
   "devDependencies": {
     "@nomiclabs/hardhat-ethers": "^2.0.5",
     "@nomiclabs/hardhat-waffle": "^2.0.3",
+    "@nomiclabs/hardhat-web3": "^2.0.0",
```
ItsNickBarry, line 16: This repository uses ethers/waffle.  It would make a mess to combine that with web3/truffle.

iam-dev: oke I will change that

iam-dev:
please add describeBehaviorOfAccessControl to @solidstate/spec

contracts/access/AccessControlInternal.sol
```diff
@@ -0,0 +1,135 @@
...
+        bytes32 indexed role,
+        address indexed account,
+        address indexed sender
+    );
+
+    event RoleRevoked(
+        bytes32 indexed role,
+        address indexed account,
+        address indexed sender
+    );
```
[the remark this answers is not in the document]
ItsNickBarry: See `IERC20Internal` for example.

ItsNickBarry:
> please add describeBehaviorOfAccessControl to @solidstate/spec

I don't think I have access to your branch.  You can add it by importing it into `spec/access/index.ts`.

ItsNickBarry (changes requested):

contracts/utils/Context.sol
```diff
@@ -0,0 +1,36 @@
...
+ * @dev Provides information about the current execution context, including the
+ * sender of the transaction and its data. While these are generally available
+ * via msg.sender and msg.data, they should not be accessed in such a direct
+ * manner, since when dealing with meta-transactions the account sending and
+ * paying for execution may not be the actual sender (as far as an application
+ * is concerned).
+ *
+ * This contract is only required for intermediate, library-like contracts.
+ */
+abstract contract Context {
```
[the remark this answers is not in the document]
iam-dev: I'm wondering what will happend if you add AccessControl to one of your Diamond Facet?

msg.sender would be the Diamond proxy itself right? and not the caller of the function

contracts/introspection/ERC165.sol
```diff
@@ -14,7 +14,13 @@ abstract contract ERC165 is IERC165 {
     /**
      * @inheritdoc IERC165
      */
-    function supportsInterface(bytes4 interfaceId) public view returns (bool) {
+    function supportsInterface(bytes4 interfaceId)
+        public
+        view
+        virtual
+        override
```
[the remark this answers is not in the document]
iam-dev: I will remove this

contracts/introspection/ERC165.sol
```diff
@@ -14,7 +14,13 @@ abstract contract ERC165 is IERC165 {
     /**
      * @inheritdoc IERC165
      */
-    function supportsInterface(bytes4 interfaceId) public view returns (bool) {
+    function supportsInterface(bytes4 interfaceId)
+        public
+        view
+        virtual
```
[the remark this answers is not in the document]
iam-dev: I will remove

contracts/utils/Context.sol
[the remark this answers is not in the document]
iam-dev: this part won't work with Proxy right? `function renounceRole(bytes32 role, address account) public virtual {
        require(
            account == msg.sender,
            'AccessControl: can only renounce roles for self'
        );

        _revokeRole(role, account);
    }`

contracts/access/AccessControl.sol
```diff
@@ -0,0 +1,139 @@
...
+ * By default, the admin role for all roles is `DEFAULT_ADMIN_ROLE`, which means
+ * that only accounts with this role will be able to grant or revoke other
+ * roles. More complex role relationships can be created by using
+ * {_setRoleAdmin}.
+ *
+ * WARNING: The `DEFAULT_ADMIN_ROLE` is also its own admin: it has permission to
+ * grant and revoke this role. Extra precautions should be taken to secure
+ * accounts that have been granted it.
+ */
+abstract contract AccessControl is Context, AccessControlInternal, ERC165 {
```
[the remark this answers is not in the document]
iam-dev: see my comment about Diamond standard

contracts/utils/Context.sol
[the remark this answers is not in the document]
ItsNickBarry: It will work.

friskyfoxdk:
Love that this is seeing progress, going to be watching this

ItsNickBarry:
I will review again soon.  Not much time at the moment.  It appears to be nearly finished, though.

contracts/access/AccessControlInternal.sol
```diff
@@ -0,0 +1,119 @@
...
+        _checkRole(role);
+        _;
+    }
+
+    /**
+     * @dev Returns `true` if `account` has been granted `role`.
+     */
+    function _hasRole(bytes32 role, address account)
+        internal
+        view
```
NouDaimon, line 36: Add virtual for overriding functionality

contracts/access/AccessControlInternal.sol
```diff
@@ -0,0 +1,119 @@
...
+        }
+    }
+
+    /**
+     * @dev Returns the admin role that controls `role`. See {grantRole} and
+     * {revokeRole}.
+     *
+     * To change a role's admin, use {_setRoleAdmin}.
+     */
+    function _getRoleAdmin(bytes32 role) internal view returns (bytes32) {
```
NouDaimon, line 81: Add virtual modifier

contracts/access/IAccessControl.sol
```diff
@@ -0,0 +1,66 @@
...
+
+pragma solidity ^0.8.0;
+
+import { IAccessControlInternal } from './IAccessControlInternal.sol';
+
+/**
+ * @dev External interface of AccessControl declared to support ERC165 detection.
+ */
+interface IAccessControl is IAccessControlInternal {
+    /**
```
NouDaimon, line 12: Follow natspec best practices for the entire file.

contracts/access/IAccessControlInternal.sol
```diff
@@ -0,0 +1,47 @@
+// SPDX-License-Identifier: MIT
+// From OpenZeppelin Contracts v4.4.1 (access/IAccessControl.sol)
+
+pragma solidity ^0.8.0;
+
+/**
+ * @title Partial AccessControl interface needed by internal functions
+ */
+interface IAccessControlInternal {
```
NouDaimon, line 9: Follow natspec best practices

contracts/utils/Strings.sol
```diff
@@ -0,0 +1,70 @@
+// SPDX-License-Identifier: MIT
+
+pragma solidity ^0.8.0;
+
+/**
+ * @dev String operations.
+ */
+library Strings {
```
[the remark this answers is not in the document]
NouDaimon: Remove this file since it will be unused.

spec/access/AccessControl.behavior.ts
```diff
@@ -0,0 +1,169 @@
...
+          .renounceRole(`${ROLE}`, authorized.address);
+
+        const trx = await instance
+          .connect(authorized)
+          .renounceRole(`${ROLE}`, authorized.address);
+        const receipt = await trx.wait();
+        expect(receipt.events?.length).to.equal(0);
+      });
+    });
+    describe('reverts if', function () {
```
NouDaimon, line 148: `describe('reverts if')` should be nested within the function they apply to; 

eg `renounceRole` should revert if `account !== msg.sender`

contracts/access/AccessControlMock.sol
```diff
@@ -0,0 +1,18 @@
...
+        _grantRole(DEFAULT_ADMIN_ROLE, msg.sender);
+        _grantRole(DEFAULT_ADMIN_ROLE, admin);
+    }
+
+    function setRoleAdmin(bytes32 roleId, bytes32 adminRoleId) public {
+        _setRoleAdmin(roleId, adminRoleId);
+    }
+
+    function senderProtected(bytes32 roleId) public onlyRole(roleId) {}
+}
```
NouDaimon, line 18: Perhaps add explicit function to test `_checkRole`?

NouDaimon (changes requested):

contracts/access/IAccessControl.sol
iam-dev: this is a copy of OpenZeppelin, which means OpenZeppelin doesn't follow natspec best practices??

ItsNickBarry: @NouDaimon SolidState hasn't settled on best practices for contract- and interface-level comments.  We'll enforce this at a later date.

@iam-dev No action needed here.

ItsNickBarry: >> OpenZeppelin doesn't follow natspec best practices??

What @NouDaimon is referring to is the "best practice" of making full use of all available tags.  OZ doesn't really do this - for example, in addition to `dev`, here they could use `title` and `author`, and maybe more.

ItsNickBarry:
@iam-dev Merge conflict fixed, pull changes.

contracts/access/IAccessControl.sol
NouDaimon: @iam-dev 'best-practices' are kind of vague, indeed; as far as I know, it is best to be as clear as possible. Therefore, as many tags as apply is the rule of thumb I go by - it is also what is done in SS :) Look at:
https://github.com/solidstate-network/solidstate-solidity/blob/master/contracts/token/ERC20/IERC20.sol

contracts/utils/Strings.sol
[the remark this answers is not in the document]
iam-dev: done

contracts/access/AccessControl.sol
```diff
@@ -0,0 +1,124 @@
...
+    function hasRole(bytes32 role, address account)
+        public
+        view
+        virtual
+        returns (bool)
+    {
+        return _hasRole(role, account);
+    }
+
+    /**
```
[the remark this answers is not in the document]
iam-dev: done

contracts/access/AccessControlMock.sol
iam-dev: done

contracts/access/AccessControl.sol
```diff
@@ -0,0 +1,124 @@
...
+     *
+     * If `account` had not been already granted `role`, emits a {RoleGranted}
+     * event.
+     *
+     * Requirements:
+     *
+     * - the caller must have ``role``'s admin role.
+     */
+    function grantRole(bytes32 role, address account)
+        public
```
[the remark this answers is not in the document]
iam-dev: solved

spec/access/AccessControl.behavior.ts
iam-dev: its the same method as Ownable.behavior.ts

NouDaimon: @iam-dev In `Ownable.behavior.ts` the revert statements are nested within the function they pertain to:
https://github.com/solidstate-network/solidstate-solidity/blob/master/spec/access/Ownable.behavior.ts

Similarly:
https://github.com/solidstate-network/solidstate-solidity/blob/9e5451f7a83a779028dc8f3ea8886fb59f1932e9/spec/token/ERC20/ERC20Base.behavior.ts#L161

contracts/access/IAccessControl.sol
ItsNickBarry: Generally I use the `title` tag to briefly describe the contract, plus a `dev` tag is there's additional miscellaneous information.  But it's only about 80% consistent across the library.  A future PR will fix, and from that point on we will enforce consistency.

iam-dev:
please check and merge

nedgar:
+1 for having this support for the access control functions!

ItsNickBarry:
Please forgive the delay, will get back to this as soon as I can.

LorranSutter:
+1 Looking forward to seeing it in the next version

jhubbardsf:
Any update on when this will be merged or is there still work that needs to be done?

ItsNickBarry:
Once again, I apologize for the delay.  I've made some changes, and I believe this is almost ready to merge.

One outstanding question: why shouldn't `grantRole` revert if a user already has a role, and why shouldn't `revokeRoke` revert if a user doesn't have a role?  I'd like to make one of two changes:
1. Revert if either of these functions is called but has no effect.
2. Add a `bool` return argument to these functions to indicate whether the call had an effect.

Any opinions?  I will merge once this is decided.

jhubbardsf:
OpenZeppelin's implementation doesn't revert on those things, so I guess that's just how it was ported. I'm wondering why they made the decision not to revert if it wouldn't do anything. Maybe there's an edge case I'm not seeing. Returning a `bool` would still give feedback but not be a "breaking" change compared to OpenZeppelin. So maybe change 2 would be my vote.

ItsNickBarry:
Yeah, the PR matches the original OZ implementation, but the same question could also be asked there.  `AccessControl` isn't based on an EIP, so it's okay to make improvements.

Thought of a third option which I'm leaning towards - see my edited comment above.

spec/access/access_control/AccessControl.behavior.ts
```diff
@@ -0,0 +1,169 @@
...
+          await instance.callStatic['getRoleAdmin(bytes32)'](
+            DEFAULT_ADMIN_ROLE,
+          ),
+        ).to.equal(DEFAULT_ADMIN_ROLE);
+      });
+    });
+
+    describe('#grantRole(bytes32,address)', function () {
+      it('adds role to account', async function () {
+        await instance.connect(admin).grantRole(ROLE, nonAdmin.address);
```
0xCourtney, line 70: The instance should already be connected to the admin signer. I noticed this in both the test files.

```suggestion
        await instance.grantRole(ROLE, nonAdmin.address);
```

0xCourtney:
For both `grantRole` and `revokeRole` checks can be performed ahead of time if a bool or revert is desirable. I like the third option or leaving it as is.

spec/access/access_control/AccessControl.behavior.ts
ItsNickBarry: Reliance on default accounts can cause unexpected behavior when spec is run in an arbitrary end-user environment.  SS is moving towards specifying sender explicitly, but there may still be some inconsistencies.

0xCourtney: Good point, ive definitely lost track of which signer is attached to which instance before so being explicit is beneficial for everyone involved.

0xCourtney (approved):

ItsNickBarry:
Implemented option 3.

test/access/AccessControl.ts
```diff
@@ -0,0 +1,106 @@
...
+        await expect(
+          instance.connect(nonAdmin).callStatic['checkRole(bytes32)'](ROLE),
+        ).to.revertedWith(
+          `AccessControl: account ${nonAdmin.address.toLowerCase()} is missing role ${ROLE}`,
+        );
+      });
+    });
+  });
+
+  describe('#_checkRole(bytes32,address)', function () {
```
NouDaimon, line 55: `_hasRole` is implicitly checked within here hence no explicit test?

NouDaimon (approved):

test/access/AccessControl.ts
ItsNickBarry: I'll add a test.

ItsNickBarry: Actually it's implicitly checked by the spec for the external version of the function.  Internal functions with 1:1 external counterparts don't need their own tests.

ItsNickBarry (approved):

Issue #98, opened 2022-05-13
Add default `ERC1155Receiver` implementation


ItsNickBarry:
A default implementation should simply accept transfer of ERC1155 tokens.

PR #99, opened 2022-05-21
Added default ERC 1155 receiver

nonfungibleayo-eth:
Added a default implementation of ERC 1155 receiver to solve this open issues 
https://github.com/solidstate-network/solidstate-solidity/issues/98


ItsNickBarry:
This won't work, because it would require us to add OpenZeppelin as a dependency in `package.json`.  We can't do this because Hardhat doesn't support nested Solidity dependencies in combination with all common package managers - I believe pnpm, for example, uses a directory structure Hardhat doesn't recognize.

What we need for issue #98 is a custom contract (or a port of the OZ version) included directly within the SolidState package.

PR #100, opened 2022-05-24
Added `ownerOf` to ERC721Enumerable

shanejonas:
`ownerOf` needs to be implemented for OpenSea and other dapps to work properly


ItsNickBarry:
This function is already defined in `ERC721Base`.  `ERC721Enumerable` is not meant to be used without `ERC721Base`.  I intend to make this more clear in upcoming documentation.

PR #115, opened 2022-06-14
implements ERC20Metadata in ERC4626Base

### LINKED ISSUE #110: ERC4626 must implement ERC20Metadata
[EIP-4626](https://eips.ethereum.org/EIPS/eip-4626) requires that ERC4626 implement the otherwise optional ERC20 metadata functions.

Attempting to setup interface and implementation inheritance yields an error:
```
TypeError: Linearization of inheritance graph impossible
```

0xCourtney:
Closes #110

When inheriting contracts its important to keep the order of inheritance cconsistent across the derived contracts. 

From the solidity docs:

> Another simplifying way to explain this is that when a function is called that is defined multiple times in different contracts, the given bases are searched from right to left (left to right in Python) in a depth-first manner, stopping at the first match. If a base contract has already been searched, it is skipped.
> ```
>...
> contract Base1 {
>     constructor() {}
> }
> 
> contract Base2 {
>     constructor() {}
> }
> 
> // Constructors are executed in the following order:
> //  1 - Base1
> //  2 - Base2
> //  3 - Derived1
> contract Derived1 is Base1, Base2 {
>     constructor() Base1() Base2() {}
> }
> ...
> ```

In this particular case `SolidStateERC4626` inherits `ERC4626Base` which has the following order:

```bash
ERC4626Base
ERC20Base
...
ERC20Metadata
IERC20Metadata // comes after ERC20Base
...
```

Followed by `SolidStateERC20`:

```bash
SolidStateERC20
...
IERC20Metadata // comes before ERC20Base
ERC20Base 
...
```

Notice that the order in which `ERC20Base` is inherited wrt `IERC20Metadata` is different between the two base contracts.


ItsNickBarry (changes requested):

ItsNickBarry:
@0xCourtney Is there a general rule for how inherited contracts should be ordered?

0xCourtney:
> @0xCourtney Is there a general rule for how inherited contracts should be ordered?

I just updated in the description.

ItsNickBarry:
From the docs:

> You have to list the direct base contracts in the order from “most base-like” to “most derived”.

My interpretation of this is that the interfaces should come first, though this PR does the opposite.

0xCourtney:
> From the docs:
> 
> > You have to list the direct base contracts in the order from “most base-like” to “most derived”.
> 
> My interpretation of this is that the interfaces should come first, though this PR does the opposite.

Good callout, since we want to conform to this, we can simply use the `IERC20Base` interface instead of `IERC20` in the `IERC4626` interface. This is much less intrusive than my previous commit.

ItsNickBarry:
Okay, I think I got to this point, but encountered the error when I tried to make `ERC4626BaseInternal` inherit from `ERC20MetadataInternal`.

0xCourtney:
> 

I think may have removed ERC20MetadataInternal on my last update. Nonetheless I may have pin-pointed the issue. I was able use Solidity Visual Developer extension to print out the inheritance chain of each contract and found that the order in which `IERC20Metadata` is inherited in `ISolidStateERC4626` is not consistent with the other contracts.

```bash
SolidStateERC4626
↖ SolidStateERC20
↖ ERC4626Base
↖ ...
↖ ISolidStateERC4626
↖ ISolidStateERC20
↖ IERC4626Base
↖ IERC4626
↖ IERC20Metadata // after IERC4626
↖ ...

ERC4626Base
↖ ...
↖ IERC4626Base
↖ IERC4626
↖ IERC20Metadata // after IERC4626
↖ ...

ISolidStateERC4626
↖ ISolidStateERC20
↖ IERC20Metadata // before IERC4626
↖ ...
↖ IERC4626
↖ ...
```

ItsNickBarry:
Closing in favor of #120.

Issue #105, opened 2022-05-30
Fix `ERC2612` interface


ItsNickBarry:
The ERC2612 interface is missing one function from the specification: https://eips.ethereum.org/EIPS/eip-2612#specification

`ERC20Permit` is a desirable feature for most tokens, and should be included in the SolidState recommended `ERC20` configuration.

ItsNickBarry:
The `eth-permit` package internally queries the `ERC20Metadata#name` function, which does should exist on the interface.

ItsNickBarry:
Because `ERC2612` is not final, it may not make sense to include it in `SolidStateERC20`.

PR #119, opened 2022-07-16
Include `ERC20Permit` in `SolidStateERC20`

ItsNickBarry:
Originally described in #105, but no longer in scope.  Merge after #118.

### DIFF
--- a/abi/IERC20BaseInternal.json
+++ b/abi/IERC20BaseInternal.json
@@ -1,0 +1,52 @@
+[
+  {
+    "anonymous": false,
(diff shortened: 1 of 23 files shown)


0xCourtney:
IIUC, the standard is to provide each internal contract with an internal interface, even if it's empty?

NouDaimon (approved):

ItsNickBarry:
@0xCourtney Maybe.  I added the empty files to make it easier to methodically check for and apply inheritance ordering, but I removed some and compilation still worked.

Issue #124, opened 2022-07-20
False pass ERC1155Base reverts if burn amount exceeds balance


ElmoArmy:
The following line contains an error.
https://github.com/solidstate-network/solidstate-solidity/blob/52c913b1e68ee5483c923690c83e611532200e37/test/token/ERC1155/ERC1155Base.ts#L310-L315

according to the contract
https://github.com/solidstate-network/solidstate-solidity/blob/52c913b1e68ee5483c923690c83e611532200e37/contracts/token/ERC1155/base/ERC1155BaseInternal.sol#L184-L187

the expected error reason should therefore be plural, ie.:
```ERC1155: burn amount exceeds balances```

---
### As to why this bug exists yet the test passes 
The ethereum-waffle chai assertion for revertedWith checks for the expected string to be anywhere in the error string. So expecting a blank space, or ```ERC```, or any other substring of the actual error would also pass this assertion and since the substring up to balance is the same, we observe this false match. 

 I'd recommend ( and so does hardhat 😆) migrating to hardhat's own [hardhat-chai-matchers](https://hardhat.org/hardhat-chai-matchers/docs/overview) which has a less footgunny exact match for this assertion.

ItsNickBarry:
Fixed.  Will be included in next release.  I'll consider switching to that alternative plugin soon.

PR #125, opened 2022-07-20
feat: migrate to Hardhat chai matchers 

ElmoArmy:
Replaces the ethereum-waffle package with @ethereum-waffle/mock-contract and @nomicfoundation/hardhat-chai-matchers.

Summary of changes:
- Removed `ethereum-waffle` and `@nomiclabs/hardhat-waffle` as dependencies
- Added  `@nomicfoundation/hardhat-chai-matchers` and  `@ethereum-waffle/mock-contract` as dependencies
- Added import of  `@nomiclabs/hardhat-ethers` hardhat configurations where the hardhat-waffle would import the hardhat-ethers plugin.
- Updated imports of deployMockContract, MockContract to come from `@ethereum-waffle/mock-contract`.
- Updated expected `Hardhat couldn't infer the reason` assertion with revertedWithoutReason matcher.
- Fixed test AddressUtils target contract reverts, with default error message assertion.

Initially introduced 4 Failing tests:
```
 4 failing

  1) SolidStateDiamond
       ::SolidStateDiamond
         ::DiamondWritable
           #diamondCut((address,enum,bytes4[])[],address,bytes)
             reverts if
               passed FacetCutAction is invalid:
     AssertionError: Expected transaction to be reverted with reason 'Hardhat couldn't infer the reason.', but it reverted without a reason


  2) DiamondWritable
       ::DiamondWritable
         #diamondCut((address,enum,bytes4[])[],address,bytes)
           reverts if
             passed FacetCutAction is invalid:
     AssertionError: Expected transaction to be reverted with reason 'Hardhat couldn't infer the reason.', but it reverted without a reason


  3) ERC1155Base
       __internal
         #_burn(address,uint256,uint256)
           reverts if
             burn amount exceeds balance:
     AssertionError: Expected transaction to be reverted with reason 'ERC1155: burn amount exceeds balance', but it reverted with reason 'ERC1155: burn amount exceeds balances'    


  4) AddressUtils
       __internal
         #functionCallWithValue(address,bytes,uint256)
           reverts if
             target contract reverts, with default error message:
     AssertionError: Expected transaction to be reverted with reason 'AddressUtils: failed low-level call', but it reverted with reason 'AddressUtils: failed low-level call with value'
```

- Failures # 1 & # 2 are occurring due to the way the new matchers work. The proper matcher would now be revertedWithoutReason.

- Failure # 3 and # 4 is related to what was brought up in  #124. # 3 was fixed upstream. # 4 is a similar issue, but in this case it is the test that has the incorrect error string.

### DIFF
--- a/spec/index.ts
+++ b/spec/index.ts
@@ -1,8 +1,3 @@
-import chai from 'chai';
-import { solidity } from 'ethereum-waffle';
-
-chai.use(solidity);
-
 export * from './access';
 export * from './factory';
 export * from './introspection';
--- a/hardhat.config.ts
+++ b/hardhat.config.ts
@@ -1,4 +1,5 @@
-import '@nomiclabs/hardhat-waffle';
+import '@nomicfoundation/hardhat-chai-matchers';
+import '@nomiclabs/hardhat-ethers';
 import '@solidstate/hardhat-4byte-uploader';
 import '@typechain/hardhat';
 import fs from 'fs';
--- a/lib/hardhat.config.ts
+++ b/lib/hardhat.config.ts
@@ -1,2 +1,1 @@
-import '@nomicfoundation/hardhat-chai-matchers';
 import '@nomiclabs/hardhat-ethers';
--- a/lib/hardhat.config.ts
+++ b/lib/hardhat.config.ts
@@ -1,1 +1,2 @@
-import '@nomiclabs/hardhat-waffle';
+import '@nomicfoundation/hardhat-chai-matchers';
+import '@nomiclabs/hardhat-ethers';
--- a/package.json
+++ b/package.json
@@ -11,8 +11,9 @@
     "upgrade-dependencies": "yarn-up -e '@solidstate/library,@solidstate/spec,@solidstate/typechain-types' && yarn upgrade"
   },
   "devDependencies": {
+    "@ethereum-waffle/mock-contract": "^3.4.4",
+    "@nomicfoundation/hardhat-chai-matchers": "^1.0.2",
     "@nomiclabs/hardhat-ethers": "^2.1.0",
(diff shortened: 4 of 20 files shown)


ItsNickBarry:
Thanks for the PR, I'll review soon.  Not much time at the moment.

spec/index.ts
```diff
@@ -1,8 +1,3 @@
-import chai from 'chai';
-import { solidity } from 'ethereum-waffle';
-
-chai.use(solidity);
```
ItsNickBarry, line 4: Not sure why this was needed.  Is it possible that removing it will break some SS installations?

ItsNickBarry (approved):
Reasonable change.  I will merge before next release.

spec/index.ts
ElmoArmy: I apologize for the delay in my response. 

Looks like @Lumyo ran into an issue in dda39ac. I didn't experience any issues building and as far as usage, it would be expected that the user of this repo's spec package would be using the `@nomicfoundation/hardhat-chai-matchers` package. hardhat-chai-matchers in turn does this setup [here](https://github.com/NomicFoundation/hardhat/blob/09ebee4aef831392ec5b5cce022d7f03fbd8c3aa/packages/hardhat-waffle/src/waffle-chai.ts).

This is therefore a breaking change for the spec package.

ItsNickBarry: I think the solution is to make the matchers package a peer dependency.  Though I haven't worked with peer dependencies much, and don't understand all of the implications.

ItsNickBarry:
This is required for PR #139, meaning that the breaking change mentioned in comments above is probably worth it.  Merging into the `error-types` branch.

PR #69, opened 2022-01-10
Add `Queue` library

ItsNickBarry:
Related to #34.

PR #129, opened 2022-07-25
Standardize and add new data structures

### LINKED ISSUE #128: Organize data structures
As we add more data structure libraries, it makes sense to standardize their interfaces and how they're published.  This will affect the scope of #126 and #127.

Data structures should all be stored in the `contracts/data/` directory.

It seems like each data structure should have an implementation for each of `address`, `uint256`, and `bytes32`.  What do you think, @0xCourtney?

ItsNickBarry:
closes #128


ItsNickBarry:
Related PRs #69, #154, and #155 still pending, but can be merged later.

PR #131, opened 2022-08-12
Add multi-initialize functionality to DiamondBase

rmcmk:
Added functionality to execute multiple initializer functions when upgrading a diamond

from:
```
https://twitter.com/mudgen/status/1557422724837957632
https://github.com/mudgen/diamond-1-hardhat/commit/f80868399dd86d1019ac979017098ad33387aed3
```


ItsNickBarry (commented):
This seems useful, but I think it should have been included as part of EIP-2535.  The new function is `internal` and not called by any other functions, so how do you envision users' integrating it into their projects?

PR #134, opened 2022-09-05
Binary Heap

### LINKED ISSUE #126: Binary Heap
The Binary Heap is commonly used for priority queues with applications in DeFi such as a DEX order book. This data structure is partially sorted, therefore only the root node is guaranteed to be in the correct order. This is particularly useful in cases where only the first entry in the queue is needed per transaction. The Binary Heap should have a generalized implementation.

0xCourtney:
Closes #126


ItsNickBarry (changes requested):
Remove `binaryheap/` directory and place contents directly in `data/`.

test/data/BinaryHeap.ts
```diff
@@ -0,0 +1,495 @@
...
+
+          expect(await instance['heap()']()).to.deep.equal([
+            twoBytes32,
+            zeroBytes32,
+            oneBytes32,
+          ]);
+        });
+      });
+
+      describe('#remove(bytes32)', () => {
```
NouDaimon, line 148: Add test for root being what it should be given the `_maxHeapify` sorting.

Also, for the nodes to be ordered in accordance with said sorting.

NouDaimon: Perhaps the latter of these are superfluous given the `#root` tests check implicitly for these. Thoughts?

test/data/BinaryHeap.ts
```diff
@@ -0,0 +1,495 @@
...
+          expect(await instance.callStatic['remove(bytes32)'](zeroBytes32)).to
+            .be.true;
+        });
+
+        it('returns false if value is not removed', async () => {
+          expect(await instance.callStatic['remove(bytes32)'](zeroBytes32)).to
+            .be.false;
+        });
+      });
+    });
```
NouDaimon, line 175: Perhaps dd explicit test for `_swap` and `_maxHeapify`?

NouDaimon (changes requested):
Minor descriptive changes only, functionality seems to check out.

Well written.

contracts/data/BinaryHeapAddressMock.sol
```diff
@@ -0,0 +1,48 @@
...
+    BinaryHeap.AddressHeap internal addressHeap;
+
+    function heap() external view returns (address[] memory) {
+        uint256 l = addressHeap.length();
+        address[] memory arr = new address[](l);
+        for (uint256 i = 0; i < l; i++) {
+            arr[i] = addressHeap.at(i);
+        }
+        return arr;
+    }
```
ItsNickBarry, line 19: Might be better to add a `toArray` function to the library itself, and then use that here.  It wouldn't be gas-efficient, but could be useful in `view` functions.  Also could probably paginate it (though I don't have an opinion about what such a function signature should look like).

ItsNickBarry (commented):
I believe `unchecked` blocks can be added in a few plaecs.  Please confirm if that's not the case.

contracts/data/BinaryHeap.sol
```diff
@@ -0,0 +1,263 @@
...
+        }
+    }
+
+    function _maxHeapify(
+        Heap storage heap,
+        uint256 l,
+        uint256 i
+    ) private {
+        uint256 largest = i;
+        uint256 lt = 2 * i + 1;
```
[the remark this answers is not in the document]
0xCourtney: agreed, being explicit doesn't hurt. I added `t` to help differentiate `l` (length) from `l` (left). This should be clearer as of [ec3c875](https://github.com/solidstate-network/solidstate-solidity/pull/134/commits/ec3c87553671ff142e11d66ab4da247d45cd4b00).

contracts/data/BinaryHeap.sol
```diff
@@ -0,0 +1,263 @@
+// SPDX-License-Identifier: MIT
+
+pragma solidity ^0.8.8;
+
+library BinaryHeap {
```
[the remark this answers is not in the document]
0xCourtney: [7227bff](https://github.com/solidstate-network/solidstate-solidity/pull/134/commits/7227bff67b360126e1ba34299a236e8031c09899)

test/data/BinaryHeap.ts
0xCourtney: > Perhaps the latter of these are superfluous given the #root tests check implicitly for these. Thoughts?

I believe this is true, sorting can only be checked if an element is removed from the tree therefore I can only check that the tree is sorted by calling something like `root` to make sure the head of the tree is the highest value. It may make sense to check that the following property holds true though `in which the value of a node is greater than or equal to the values of its children`.

NouDaimon: If that property is satisfied, then for sure the algorithm is working correctly and all above suggestions in this comment thread (root check eg) are needless.

0xCourtney:
> I believe `unchecked` blocks can be added in a few plaecs. Please confirm if that's not the case.

I added a few unchecked blocks in [877e2b3](https://github.com/solidstate-network/solidstate-solidity/pull/134/commits/877e2b3c53d36c103b56aa2c1f78dab1907d0582)

contracts/data/BinaryHeapAddressMock.sol
0xCourtney: added in [ca1de73](https://github.com/solidstate-network/solidstate-solidity/pull/134/commits/ca1de730c91abf77147a36748336aeaa0fd5d169)

test/data/BinaryHeap.ts
0xCourtney: @NouDaimon I think you've touched on something that I haven't seen so far in SS. `_swap` and `_maxHeapify` are both private functions that are not exposed externally like the other functions in the `BinaryHeap` library. This presents a unique challenge for writing tests. I think there are three possible solutions here, keep in mind, this situation may come up again so this could be an excellent time to have a plan for dealing with it.

1. Keep `_swap` and `_maxHeapify` private but copy-paste their logic to the mock binary heap files.
2. Set `_swap` and `_maxHeapify` as internal functions
3. Don't test `_swap` and `_maxHeapify` directly

Would certainly be interested in hearing your thoughts on this. I'm torn between 2 and 3, on one hand, I think it makes sense to test as much as possible but adding these functions to the interface doesn't sit well with me. On the other hand, a robust set of tests may make it unnecessary to test each and every function, but this makes it challenging to track regressions and cover edge cases.

cc @ItsNickBarry

NouDaimon: I vote for 3, in the cases of `private` functions - `internal` functions tend to have functions exposing them and can be tested in that manner.

Every function that invokes `private` functions should test that the `private` functions work as expected, since they are now part of the function being tested. As a result, the test for each function invoking `private` functions has to be more robust - this will aid in catching edge cases as well, should the `private` functionality change.

The goal with the above practice is to make tests as self-contained as possible.

Using the `BinaryHeap` as an example, the tests for `_remove` should check that the root of the remaining heap is indeed node with the largest (since `_maxHeapify`) value - this ensure that this specific result of the `_maxHeapify` function which is invoked in `_remove` works as expected.

ItsNickBarry: No need to test `private` functions.  All behavior of such functions should be testable via the `internal` or `external` functions.

test/data/BinaryHeap.ts
```diff
@@ -0,0 +1,740 @@
...
+          await instance['add(bytes32)'](eightBytes32);
+          await instance['add(bytes32)'](threeBytes32);
+          await instance['add(bytes32)'](sixBytes32);
+          await instance['add(bytes32)'](fourBytes32);
+          await instance['add(bytes32)'](fiveBytes32);
+          await instance['add(bytes32)'](nineBytes32);
+          await instance['add(bytes32)'](sevenBytes32);
+          await instance['add(bytes32)'](oneBytes32);
+
+          await instance['add(bytes32)'](sixBytes32);
```
NouDaimon, line 211: Should this be:
```suggestion
          await instance['remove(bytes32)'](sixBytes32);
```
?

test/data/BinaryHeap.ts
```diff
@@ -0,0 +1,740 @@
...
+          checkNodes(await instance['toArray()']());
+
+          await instance['remove(bytes32)'](fiveBytes32);
+          checkNodes(await instance['toArray()']());
+
+          await instance['add(bytes32)'](eightBytes32);
```
NouDaimon, line 217: Should this be:

```suggestion
          await instance['remove(bytes32)'](eightBytes32);
```
?

test/data/BinaryHeap.ts
```diff
@@ -0,0 +1,740 @@
...
+          await instance['add(bytes32)'](twoBytes32);
+
+          expect(await instance.callStatic['remove(bytes32)'](zeroBytes32)).to
+            .be.true;
+        });
+
+        it('returns false if value does not exist', async () => {
+          expect(await instance.callStatic['remove(bytes32)'](zeroBytes32)).to
+            .be.false;
+        });
```
NouDaimon, line 254: Up to you whether to add a test which checks that the highest value is the root, when an element is removed.

I guess this is included in `checkNodes` although explicit root checking may be worthwhile.

test/data/BinaryHeap.ts
```diff
@@ -0,0 +1,740 @@
...
+          await instance['add(address)'](eightAddress);
+          await instance['add(address)'](threeAddress);
+          await instance['add(address)'](sixAddress);
+          await instance['add(address)'](fourAddress);
+          await instance['add(address)'](fiveAddress);
+          await instance['add(address)'](nineAddress);
+          await instance['add(address)'](sevenAddress);
+          await instance['add(address)'](oneAddress);
+
+          await instance['add(address)'](sixAddress);
```
NouDaimon, line 448: Should this be:
```suggestion
          await instance['remove(address)'](sixAddress);
```
?

test/data/BinaryHeap.ts
```diff
@@ -0,0 +1,740 @@
...
+          await instance['add(address)'](twoAddress);
+
+          expect(await instance.callStatic['remove(address)'](zeroAddress)).to
+            .be.true;
+        });
+
+        it('returns false if value does not exist', async () => {
+          expect(await instance.callStatic['remove(address)'](zeroAddress)).to
+            .be.false;
+        });
```
NouDaimon, line 491: Similar to comment for `Bytes32Heap` for root test.

test/data/BinaryHeap.ts
```diff
@@ -0,0 +1,740 @@
...
+          await instance['add(uint256)'](eight);
+          await instance['add(uint256)'](three);
+          await instance['add(uint256)'](six);
+          await instance['add(uint256)'](four);
+          await instance['add(uint256)'](five);
+          await instance['add(uint256)'](nine);
+          await instance['add(uint256)'](seven);
+          await instance['add(uint256)'](one);
+
+          await instance['add(uint256)'](six);
```
NouDaimon, line 684:
```suggestion
          await instance['remove(uint256)'](six);
```

NouDaimon (changes requested):

test/data/BinaryHeap.ts
0xCourtney: added in [76247ee6](https://github.com/solidstate-network/solidstate-solidity/pull/134/commits/76247ee61179a71413f271013c39573ef116b430)

test/data/BinaryHeap.ts
0xCourtney: Good catch! Fixed in https://github.com/solidstate-network/solidstate-solidity/pull/134/commits/d107a575b860d1a6784ec410650506829d51e276

test/data/BinaryHeap.ts
0xCourtney: added in https://github.com/solidstate-network/solidstate-solidity/pull/134/commits/0d02343d3377301915ef9617f5780e79420e3032

NouDaimon (approved):

contracts/data/BinaryHeap.sol
```diff
@@ -0,0 +1,320 @@
...
+        Heap storage heap,
+        uint256 a,
+        uint256 b
+    ) private {
+        (heap._indexes[heap._values[a]], heap._indexes[heap._values[b]]) = (
+            heap._indexes[heap._values[b]],
+            heap._indexes[heap._values[a]]
+        );
+
+        (heap._values[a], heap._values[b]) = (heap._values[b], heap._values[a]);
```
ItsNickBarry, line 318: I believe this could be made more efficient by storing a local reference to `_indexes` and `_values`, rather than accessing via `heap` each time.

ItsNickBarry: Same might be true for `_values` in the `_maxHeapify` function above.

contracts/data/BinaryHeap.sol
```diff
@@ -0,0 +1,320 @@
...
+    function _maxHeapify(
+        Heap storage heap,
+        uint256 len,
+        uint256 index
+    ) private {
+        uint256 largest = index;
+
+        unchecked {
+            uint256 left = 2 * index + 1;
+            uint256 right = 2 * index + 2;
```
ItsNickBarry, line 290:
```suggestion
            uint256 right = left + 1;
```

contracts/data/BinaryHeap.sol
```diff
@@ -0,0 +1,320 @@
...
+    }
+
+    function _heapify(Heap storage heap) private {
+        uint256 len = _length(heap);
+        if (len > 1) {
+            unchecked {
+                uint256 index = ((len) / 2) - 1;
+                while (index >= 0) {
+                    _maxHeapify(heap, len, index);
+                    if (index == 0) break;
```
ItsNickBarry, line 274: Why would `index == 0`?  That would fail the `while` condition.

contracts/data/BinaryHeap.sol
```diff
@@ -0,0 +1,320 @@
...
+                    --index;
+                }
```
ItsNickBarry, line 276: This would be better expressed as a `for` loop than a `while` loop.

contracts/data/BinaryHeap.sol
```diff
@@ -0,0 +1,320 @@
...
+            return false;
+        }
+    }
+
+    function _remove(Heap storage heap, bytes32 value) private returns (bool) {
+        if (_contains(heap, value)) {
+            uint256 index = _indexOf(heap, value);
+
+            // move node with last element in the tree, then remove it
+            _swap(heap, index, _length(heap) - 1);
```
ItsNickBarry, line 254: I believe this can be `unchecked` because if the `_contains` check passes, the `_length` must be at least 1.

contracts/data/BinaryHeap.sol
```diff
@@ -0,0 +1,320 @@
...
+        view
+        returns (uint256[] memory)
+    {
+        uint256 len = _length(heap._inner);
+        uint256[] memory arr = new uint256[](len);
+        for (uint256 index = 0; index < len; index++) {
+            arr[index] = uint256(_at(heap._inner, index));
+        }
+        return arr;
+    }
```
ItsNickBarry, line 200: The `for` loops in each of the `toArray` functions can be marked `unchecked`.

ItsNickBarry: Also, instead of casting the value returned by the low-level `_at` function, why not use the higher-level `at` functions that already return the desired type?

ItsNickBarry (changes requested):

contracts/data/BinaryHeap.sol
ItsNickBarry: Oh, I see.  I think this is a case for a `do / while` loop:
```solidity
do {
  ...
} while (index > 0);
```

contracts/data/BinaryHeap.sol
0xCourtney: Great suggestion! https://github.com/solidstate-network/solidstate-solidity/pull/134/commits/2f4fde36bcf8cd3220fd89abdf6972aef5263be4

contracts/data/BinaryHeap.sol
0xCourtney: https://github.com/solidstate-network/solidstate-solidity/pull/134/commits/2f4fde36bcf8cd3220fd89abdf6972aef5263be4

contracts/data/BinaryHeap.sol
ItsNickBarry: Right, never mind.

contracts/data/BinaryHeap.sol
0xCourtney: > I believe this could be made more efficient by storing a local reference to `_indexes` and `_values`, rather than accessing via `heap` each time.

this showed substantial gas savings https://github.com/solidstate-network/solidstate-solidity/pull/134/commits/df978f245df99fa3073d68abfb9aabb4e8dc1d89

0xCourtney: > Same might be true for `_values` in the `_maxHeapify` function above.

https://github.com/solidstate-network/solidstate-solidity/pull/134/commits/b9e9dc14ed0f76a2aaae6d3e23e7c84bcb941415

contracts/data/BinaryHeap.sol
```diff
@@ -0,0 +1,340 @@
...
+    function _swap(
+        Heap storage heap,
+        uint256 a,
+        uint256 b
+    ) private {
+        bytes32[] storage values = heap._values;
+        bytes32 aValue = values[a];
+        bytes32 bValue = values[b];
+
+        (heap._indexes[aValue], heap._indexes[bValue]) = (
```
ItsNickBarry, line 333: Now try this:
```solidity
mapping(bytes32 => uint256) storage indexes = heap._indexes;
...
```

0xCourtney: Am I missing something, this seems to increase gas costs.

```js
function _swap(
    Heap storage heap,
    uint256 a,
    uint256 b
) private {
    bytes32[] storage values = heap._values;
    mapping(bytes32 => uint256) storage indexes = heap._indexes;

    bytes32 aValue = values[a];
    bytes32 bValue = values[b];

    (indexes[aValue], indexes[bValue]) = (indexes[bValue], indexes[aValue]);
    (values[a], values[b]) = (bValue, aValue);
}
```
Before
Screen Shot 2022-09-29 at 14 58 50

After
Screen Shot 2022-09-29 at 14 59 10

ItsNickBarry: Maybe do all of the `values` logic first, then the `indexes` logic after?  If that doesn't work, just ignore.  Sometimes these ridiculous optimizations are unpredictable.

ItsNickBarry: ^ that probably won't work, but I'm curious.

0xCourtney: I'm very curious to know why this changes gas costs

```js
function _swap(
    Heap storage heap,
    uint256 a,
    uint256 b
) private {
    bytes32[] storage values = heap._values;

    bytes32 aValue = values[a];
    bytes32 bValue = values[b];

    (values[a], values[b]) = (bValue, aValue);

    mapping(bytes32 => uint256) storage indexes = heap._indexes;
    (indexes[aValue], indexes[bValue]) = (indexes[bValue], indexes[aValue]);
}
```

Screen Shot 2022-09-29 at 15 21 31

0xCourtney: https://github.com/solidstate-network/solidstate-solidity/pull/134/commits/5182760f94b5f7d732f633c31ac243a16e616b09

ItsNickBarry: Yeah, I was thinking it would require less stack reorganization.

0xCourtney:
closing in favor of #143

PR #136, opened 2022-09-14
Library to handle Merkle trees on-chain

NouDaimon:
Contains:
- `MerkleTree` struct
- `updateNode` function to update a leaf node hash and the associated parent hashes 

Potential changes:
- Make `updateNode` private
- Add `remove`, `update` and `add` functions, which use `updateNode`


ItsNickBarry (commented):
@NouDaimon Added an alternative implementation.  It's similar to yours, but I think it's simpler.  Also added some comments.

contracts/data/IncrementalMerkleTree2.sol
```diff
@@ -0,0 +1,104 @@
...
+
+pragma solidity ^0.8.0;
+
+import 'hardhat/console.sol';
+
+library IncrementalMerkleTree2 {
+    using IncrementalMerkleTree2 for Tree;
+
+    struct Tree {
+        bytes32[][] nodes;
```
ItsNickBarry, line 11: @NouDaimon I think the main disadvantage of this approach is the lack of support an `indexOf` function.  However, I don't think it's a fundamental part of the data structure, and end users could add their own implementations if they want to spend the gas.

NouDaimon:
Several implementations of Merkle trees always mantain a balanced tree - from what I have seen, this is either done via initialization or duplication. 

 An example is the `Progressive Merkle Tree` for the new ETH deposit contract  (denoted in the image [here](https://www.researchgate.net/figure/Illustration-of-the-incremental-Merkle-tree-algorithm-Node-numbers-are-labeled-in-the_fig1_342969698))which initializes all leaves as `hash(0)` and calculates the root based on these hashes, as is normally done in Merkle Trees.
 
 Another example is the `Bitcoin` implementation (denoted[ here](https://medium.com/coinmonks/merkle-trees-concepts-and-use-cases-5da873702318)) which duplicates the leaf node, should a pair not exist, and hashes upwards to calculate the root as normal.
 
 The final example for `Monero` (again denoted [here](https://medium.com/coinmonks/merkle-trees-concepts-and-use-cases-5da873702318)) seems to follow the same implementation as the one in the SolidState `IncrementalMerkleTree.sol` library; the appended node is bubbled upwards to calculate a new root from itself and the old root. 
 
 Are there any security considerations for not initializing the `MerkleTree` or growing it in this fashion? Any considerations/potential attack vectors from using a 'dynamic height' and unbalance tree?

test/data/IncrementalMerkleTree.ts
```diff
@@ -121,6 +121,24 @@ describe('IncrementalMerkleTree', function () {
 
       expect(await instance.callStatic.root()).to.equal(hash);
     });
+
+    it('POC: returns Merkle root derived from elements contained in unbalanced tree', async () => {
```
ItsNickBarry, line 125: What is "POC"?

ItsNickBarry: This test seems a bit redundant.  Though, there's a lot of overlap in several of the tests.

ItsNickBarry:
Need confirmation that the improved hashing function will not cause issues for end-users who might be working with data in memory.

Otherwise, I think this is complete.

NouDaimon:
> Need confirmation that the improved hashing function will not cause issues for end-users who might be working with data in memory.
> 
> Otherwise, I think this is complete.

As per the discussion, given the [docs](https://docs.soliditylang.org/en/v0.8.17/internals/layout_in_memory.html?highlight=free%20memory%20slot#layout-in-memory), the memory space being used in the hash function is "scratch space" so should not affect end users memory.

Auditor should confirm.

ItsNickBarry:
I updated so that the tree accepts hashes directly, rather than raw data.  The end user must do the hashing.

hinterlist (approved):

ItsNickBarry:
@NouDaimon I'm merging this now.  We can merge the optimizations later, after further review or audit.

Thanks @hinterlist.

Issue #107, opened 2022-06-03
Add `Initializable` contract


ItsNickBarry:
Simple contract that exposes a modifier to prevent calling an `initialize` function more than once.

PR #142, opened 2022-10-02
feat(initializable): add initializable

wiasliaw:
#107


ItsNickBarry (changes requested):
Partial review.  More later.

contracts/security/Initializable/Initializable.sol
```diff
@@ -0,0 +1,7 @@
+// SPDX-License-Identifier: MIT
+
+pragma solidity ^0.8.8;
+
+import { InitializableInternal } from './InitializableInternal.sol';
+
+abstract contract Initializable is InitializableInternal {}
```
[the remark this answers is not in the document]
wiasliaw: I didn't design any external function for two reason.

1. User might design their own init function.
  - Personally, I don't like the `bytes memory` as parameter due to that user need to handle the bytes data by themselves, like `address owner = abi.decode(data, (address))`. It is ease to cause false setup.  And I think `initializer` is enough.
2. To provider well contract name to inherit.

i.e.
```solidity
// better
contract Simple is Initializable { ... }

// bad
contract Simple is InitializableInternal { ... }
```

contracts/security/Initializable/InitializableInternal.sol
```diff
@@ -0,0 +1,74 @@
...
+    using AddressUtils for address;
+    using InitializableStorage for InitializableStorage.Layout;
+
+    event Initialized(uint8 version);
+
+    error AlreadyInitialized();
+
+    error NotInitializing();
+
+    error IsInitializing();
```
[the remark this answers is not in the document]
wiasliaw: Fixed.

contracts/security/Initializable/InitializableStorage.sol
```diff
@@ -0,0 +1,20 @@
+// SPDX-License-Identifier: MIT
+
+pragma solidity ^0.8.8;
+
+library InitializableStorage {
+    struct Layout {
+        uint8 initialized;
```
ItsNickBarry, line 7: I think it would be more gas-efficient to use a `uint256` - smaller uint types generally are only more efficient when they can be packed in storage.  I believe the `bool initializing` uses a full 32-bytes storage slot, so packing does not occur.

ItsNickBarry: Though, if `initializing` were converted to something like a `bytes1`, then I believe packing would occur.

contracts/security/Initializable/Initializable.sol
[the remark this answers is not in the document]
ItsNickBarry: I'm still not sure, but for now the current API is acceptable.  If I decide to change this, I will make a new PR in the future.

contracts/security/Initializable/InitializableInternal.sol
```diff
@@ -0,0 +1,74 @@
...
+
+    error Initializable__NotInitializing();
+
+    error Initializable__IsInitializing();
+
+    modifier initializer() {
+        InitializableStorage.Layout storage l = InitializableStorage.layout();
+        bool isTopLevelCall = !l.initializing;
+        bool passed = (isTopLevelCall && l.initialized < 1) ||
+            (!address(this).isContract() && l.initialized == 1);
```
ItsNickBarry, line 24: These variable declarations use 4 of the 16 available stack positions, which adds limitations to any function that uses the modifier.  `passed` can be removed easily, because it's only used once.  `isTopLevelCall` can be replaced with direct lookups to `!l.initializing` - this will cost more gas, but that's not a significant concern for an initializer function, which is called very rarely.

contracts/security/Initializable/InitializableInternal.sol
```diff
@@ -0,0 +1,74 @@
...
+        }
+    }
+
+    modifier reinitializer(uint8 version) {
+        InitializableStorage.Layout storage l = InitializableStorage.layout();
+        // require(
+        //     !l.initializing && l.initialized < version,
+        //     'Initializable: contract is already initialized'
+        // );
+        bool passed = !l.initializing && l.initialized < version;
```
ItsNickBarry, line 43: See above comment.

contracts/security/Initializable/InitializableInternal.sol
```diff
@@ -0,0 +1,74 @@
...
+        if (!passed) revert Initializable__AlreadyInitialized();
+        l.initialized = version;
+        l.initializing = true;
+        _;
+        l.initializing = false;
+        emit Initialized(version);
+    }
+
+    modifier onlyInitializing() {
+        if (!InitializableStorage.layout().initializing)
```
ItsNickBarry, line 53:
```suggestion
        if (!_isInitializing())
```

ItsNickBarry (changes requested):
A couple of the new directories need to be renamed (change to lower case):
* `contracts/security/Initializable` => `contracts/security/initializable`
* `contracts/security/Pausable` => `contracts/security/pausable`

contracts/security/Initializable/InitializableStorage.sol
wiasliaw: Nope, uint8 and bool will be packed into a single slot. `uint8` and `bool` are both 1 byte.
I think uint8 can change to uint248 to full a slot.

BTW, `bool` uses a full slot only when `mapping(uint256 => bool)`, due to how mapping decide which slot to store.

contracts/security/Initializable/InitializableInternal.sol
wiasliaw: 1. I removed `passed`.
2. I think that `isTopLevelCall` is meaningful to know the init function whether is the first call or not.

contracts/security/Initializable/InitializableInternal.sol
wiasliaw: I removed passed.

contracts/security/Initializable/InitializableInternal.sol
wiasliaw: Resolved.

contracts/security/Initializable/InitializableStorage.sol
ItsNickBarry: I must have misunderstood - I read somewhere that `bool` takes more space than one would expect, but that presumed expectation must have been a single bit.

wiasliaw:
@ItsNickBarry IDK why renamed folders are counted in the last commit. Renamed is done.

ItsNickBarry:
The code from this PR has been included in #199 and modified in such a way that it does not work as originally intended.  Published in `v0.0.55`.  Thanks @wiasliaw for the contribution.

PR #143, opened 2022-10-04
Binary Heap

### LINKED ISSUE #126: Binary Heap

0xCourtney:
closes #126


0xCourtney:
The `require` statement should be changed to revert, I will do this after #139 is merged.

test/data/BinaryHeap.ts
```diff
@@ -0,0 +1,839 @@
...
+          checkNodes(await instance['toArray()']());
+        });
+
+        it('returns true if value is added', async () => {
+          expect(await instance.callStatic['add(bytes32)'](zeroBytes32)).to.be
+            .true;
+        });
+
+        it('returns false if value has already been added', async () => {
+          await instance.callStatic['remove(bytes32)'](zeroBytes32);
```
NouDaimon, line 201:
```suggestion
          await instance['add(bytes32)'](zeroBytes32);
```

test/data/BinaryHeap.ts
```diff
@@ -0,0 +1,839 @@
...
+          expect(await instance.callStatic['remove(bytes32)'](zeroBytes32)).to
```
NouDaimon, line 202:
```suggestion
          expect(await instance.callStatic['add(bytes32)'](zeroBytes32)).to
```

test/data/BinaryHeap.ts
```diff
@@ -0,0 +1,839 @@
...
+          checkNodes(await instance['toArray()']());
+        });
+
+        it('returns true if value is added', async () => {
+          expect(await instance.callStatic['add(address)'](zeroAddress)).to.be
+            .true;
+        });
+
+        it('returns false if value has already been added', async () => {
+          await instance.callStatic['remove(address)'](zeroAddress);
```
NouDaimon, line 471:
```suggestion
          await instance['add(address)'](zeroAddress);
```

test/data/BinaryHeap.ts
```diff
@@ -0,0 +1,839 @@
...
+          expect(await instance.callStatic['remove(address)'](zeroAddress)).to
```
NouDaimon, line 472:
```suggestion
          expect(await instance.callStatic['add(address)'](zeroAddress)).to
```

test/data/BinaryHeap.ts
```diff
@@ -0,0 +1,839 @@
...
+          await instance['add(uint256)'](one);
+          checkNodes(await instance['toArray()']());
+        });
+
+        it('returns true if value is added', async () => {
+          expect(await instance.callStatic['add(uint256)'](zero)).to.be.true;
+        });
+
+        it('returns false if value has already been added', async () => {
+          await instance.callStatic['remove(uint256)'](zero);
```
NouDaimon, line 740:
```suggestion
          await instance['add(uint256)'](zero);
```

test/data/BinaryHeap.ts
```diff
@@ -0,0 +1,839 @@
...
+          expect(await instance.callStatic['remove(uint256)'](zero)).to.be
```
NouDaimon, line 741:
```suggestion
          expect(await instance.callStatic['add(uint256)'](zero)).to.be
```

NouDaimon (approved):

contracts/data/BinaryHeap.sol
```diff
@@ -0,0 +1,341 @@
+// SPDX-License-Identifier: MIT
+
+pragma solidity ^0.8.8;
+
+/**
+ * @title Binary Heap implementation
+ * @dev The data strucutre is configured as a max-heap
```
ItsNickBarry, line 7:
```suggestion
 * @dev The data structure is configured as a max-heap
```

contracts/data/BinaryHeap.sol
```diff
@@ -0,0 +1,339 @@
...
+
+pragma solidity ^0.8.8;
+
+/**
+ * @title Binary Heap implementation
+ * @dev The data structure is configured as a max-heap
+ */
+
+library BinaryHeap {
+    error BinaryHeap__IndexOutOfBounds();
```
ItsNickBarry, line 11: I think we can remove this error and fall back to the default panic opcode.

ItsNickBarry (commented):
@0xCourtney I made a few optimizations that you should look at.  Let me know if you agree.

contracts/data/BinaryHeap.sol
```diff
@@ -304,7 +304,7 @@ library BinaryHeap {
         bytes32[] storage values = heap._values;
 
         unchecked {
-            uint256 left = 2 * index + 1;
+            uint256 left = (index << 1) | 1;
```
0xCourtney, line 301: 🔥

contracts/data/BinaryHeap.sol
0xCourtney: https://github.com/solidstate-network/solidstate-solidity/pull/143/commits/c692368e9a52ba3a541db97b4d8fe1966192d2a6

ItsNickBarry (approved):
@0xCourtney I made some significant changes to the tests.  Random values are used  instead of the precalculated constants.  This doesn't improve the integrity of the tests, but they're now a bit more concise.

Merge if you're okay with these changes.

ItsNickBarry:
Actually, I'm going to merge now so that the `data-structures` branches can be consolidated and published.  Comment if you have issues with the test updates.

PR #144, opened 2022-10-04
Update and test `EnumerableSet`

0xCourtney:


test/data/EnumerableSet.ts
```diff
@@ -0,0 +1,195 @@
...
+          await instance['add(bytes32)'](twoBytes32);
+
+          expect(await instance['at(uint256)'](0)).to.equal(zeroBytes32);
+          expect(await instance['at(uint256)'](1)).to.equal(oneBytes32);
+          expect(await instance['at(uint256)'](2)).to.equal(twoBytes32);
+        });
+
+        describe('reverts if', function () {
+          it('index out of bounds', async () => {
+            await expect(instance['at(uint256)'](0)).to.be.revertedWith(
```
NouDaimon, line 38: Perhaps wait until #139 is merged before continuing with tests?

NouDaimon: Actually, this is minor and can be retroactively fixed.

test/data/EnumerableSet.ts
```diff
@@ -0,0 +1,195 @@
...
+          await instance['remove(bytes32)'](zeroBytes32);
+          expect(await instance['length()']()).to.be.equal(0);
+          expect(await instance['indexOf(bytes32)'](zeroBytes32)).to.be.equal(
+            ethers.constants.MaxUint256,
+          );
+        });
+
+        describe('#toArray()', () => {
+          it('returns the set as an array', async () => {
+            await instance['add(bytes32)'](zeroBytes32);
```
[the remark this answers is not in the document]
ItsNickBarry: Are you talking about using a different matcher?  Can't remember which one - maybe `to.deep.have.members` or something.

test/data/EnumerableSet.ts
ItsNickBarry: Merged in that branch, can be fixed now.

test/data/EnumerableSet.ts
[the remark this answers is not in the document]
ItsNickBarry: I agree with that in theory, but it only really matters when writing spec.

test/data/EnumerableSet.ts
```diff
@@ -0,0 +1,195 @@
...
+    let instance: EnumerableSetBytes32Mock;
+    let deployer: SignerWithAddress;
+
+    beforeEach(async () => {
+      [deployer] = await ethers.getSigners();
+      instance = await new EnumerableSetBytes32Mock__factory(deployer).deploy();
+    });
+
+    describe('__internal', () => {
+      const zeroBytes32 = bnToBytes32(ethers.constants.Zero);
```
ItsNickBarry, line 21: There's also `ethers.constants.HashZero`, but your way is probably more clear.

test/data/EnumerableSet.ts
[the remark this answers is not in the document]
ItsNickBarry: Never mind, you're talking about adding the elements in a different order and testing that they're output in sorted order, correct?  In that case, yes, mix up the insertion order and add a comment explaining why.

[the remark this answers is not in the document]
ItsNickBarry: Do the same for the `at` test.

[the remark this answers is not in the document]
ItsNickBarry: Deleted some previous comments - forgot which PR I was reviewing.

test/data/EnumerableSet.ts
```diff
@@ -0,0 +1,195 @@
...
+          expect(await instance.callStatic['remove(bytes32)'](zeroBytes32)).to
+            .be.true;
+        });
+
+        it('returns false if value is not removed', async () => {
+          expect(await instance.callStatic['remove(bytes32)'](zeroBytes32)).to
+            .be.false;
+        });
+
+        it('removes value from index mapping and array', async () => {
```
ItsNickBarry, line 155: I would probably move these tests to the `length` and `indexOf` blocks.  Seems like too many things are happening here.  I don't feel strongly about it, however.

ItsNickBarry (commented):
Tests seem good, okay to make copies for `AddressSet` and `UintSet`.

test/data/EnumerableSet.ts
NouDaimon: I think the purpose is to test all effects of the function in an effort for self-containment.

The remove function explicitly alters these properties.

ItsNickBarry: If we separate test concerns perfectly, the `remove` tests must test the side effects to `length` and `indexOf`, and both the `length` and `indexOf` tests must test what happens when `remove` is called.  This is a bit too redundant, so we must choose which approach makes more sense.  There's no right answer.

0xCourtney: I removed the "removes value from index mapping and array" test case. The array length is already checked when values are added/ removed from the set in the "returns length of enumerable set" test case. I modified the test case "returns index of the value" to check that the mapping is updated when values are added/ removed from the set.

https://github.com/solidstate-network/solidstate-solidity/pull/144/commits/ffb9354f2cb3b9c6855cf7689266180c7c95d3c8

test/data/EnumerableSet.ts
0xCourtney: Awesome, I left it as is for consistency but I will be sure to keep this in mind when I need a byte32 sized zero hash.

test/data/EnumerableSet.ts
0xCourtney: I will also update this test case in the BinaryHeap PR, as its checked there as well.

NouDaimon (approved):

ItsNickBarry (approved):
Okay to merge, but not until after #143.

contracts/data/EnumerableSet.sol
```diff
@@ -150,6 +150,57 @@ library EnumerableSet {
...
+    function toArray(UintSet storage set)
+        internal
+        view
+        returns (uint256[] memory)
+    {
+        uint256 len = _length(set._inner);
+        uint256[] memory arr = new uint256[](len);
+
+        unchecked {
+            for (uint256 index = 0; index < len; index++) {
```
NouDaimon, line 196: Should look like:
```
for (uint256 index; index < len;) {
      arr[index] = at(set, index);
      ++i;
}
```

No need to initialize the index.

NouDaimon: Same applies for other for loops.

0xCourtney: https://github.com/solidstate-network/solidstate-solidity/pull/144/commits/f83e3986fb7d6d767a7130e61e0eef3842296c15

NouDaimon (approved):

Issue #92, opened 2022-05-02
Replace usage of mock contracts with `hardhat-exposed` plugin


ItsNickBarry:
https://www.npmjs.com/package/hardhat-exposed

PR #147, opened 2022-10-08
ERC2981

### LINKED ISSUE #140: Add `ERC2981` Implementation
https://eips.ethereum.org/EIPS/eip-2981

0xCourtney:
closes #140

### DIFF
--- a/contracts/interfaces/IERC2981.sol
+++ b/contracts/interfaces/IERC2981.sol
@@ -11,11 +11,11 @@ import { IERC2981Internal } from './IERC2981Internal.sol';
  */
 interface IERC2981 is IERC2981Internal, IERC165 {
     /**
-     * @notice called with the sale price to determine how much royalty is owed and to whom.
-     * @param tokenId the NFT asset queried for royalty information
-     * @param salePrice the sale price of the NFT asset specified by tokenId
-     * @return receiever address of who should be sent the royalty payment
-     * @return royaltyAmount the royalty payment amount for salePrice
+     * @notice called with the sale price to determine how much royalty is owed and to whom
+     * @param tokenId the ERC721 or ERC1155 token id to query for royalty information
+     * @param salePrice the sale price of the given asset
+     * @return receiever rightful recipient of royalty
+     * @return royaltyAmount amount of royalty owed
      */
     function royaltyInfo(uint256 tokenId, uint256 salePrice)
         external
--- a/contracts/interfaces/IERC2981.sol
+++ b/contracts/interfaces/IERC2981.sol
@@ -2,8 +2,8 @@
 
 pragma solidity ^0.8.8;
 
-import { IERC2981Internal } from './IERC2981Internal.sol';
 import { IERC165 } from './IERC165.sol';
+import { IERC2981Internal } from './IERC2981Internal.sol';
 
 /**
  * @title ERC2981 interface
--- a/contracts/interfaces/IERC2981.sol
+++ b/contracts/interfaces/IERC2981.sol
@@ -1,0 +1,23 @@
+// SPDX-License-Identifier: MIT
+
+pragma solidity ^0.8.8;
+
+import { IERC165 } from './IERC165.sol';
+
+/**
+ * @title ERC2981 interface
+ * @dev see https://eips.ethereum.org/EIPS/eip-2981
+ */
+interface IERC2981 is IERC165 {
+    /**
+     * @notice called with the sale price to determine how much royalty is owed and to whom.
+     * @param tokenId the NFT asset queried for royalty information
+     * @param salePrice the sale price of the NFT asset specified by tokenId
+     * @return address of who should be sent the royalty payment
+     * @return the royalty payment amount for salePrice
+     */
+    function royaltyInfo(uint256 tokenId, uint256 salePrice)
+        external
+        view
+        returns (address, uint256);
+}
--- a/contracts/interfaces/IERC2981.sol
+++ b/contracts/interfaces/IERC2981.sol
@@ -2,22 +2,23 @@
 
 pragma solidity ^0.8.8;
 
+import { IERC2981Internal } from './IERC2981Internal.sol';
 import { IERC165 } from './IERC165.sol';
 
 /**
  * @title ERC2981 interface
  * @dev see https://eips.ethereum.org/EIPS/eip-2981
  */
-interface IERC2981 is IERC165 {
+interface IERC2981 is IERC2981Internal, IERC165 {
     /**
      * @notice called with the sale price to determine how much royalty is owed and to whom.
      * @param tokenId the NFT asset queried for royalty information
      * @param salePrice the sale price of the NFT asset specified by tokenId
-     * @return address of who should be sent the royalty payment
-     * @return the royalty payment amount for salePrice
+     * @return receiever address of who should be sent the royalty payment
+     * @return royaltyAmount the royalty payment amount for salePrice
      */
     function royaltyInfo(uint256 tokenId, uint256 salePrice)
         external
         view
-        returns (address, uint256);
+        returns (address receiever, uint256 royaltyAmount);
 }
--- a/contracts/interfaces/IERC2981Internal.sol
+++ b/contracts/interfaces/IERC2981Internal.sol
@@ -2,12 +2,9 @@
 
 pragma solidity ^0.8.8;
 
-import { IERC2981 } from '../../../interfaces/IERC2981.sol';
-
 /**
  * @title ERC2981 interface
  */
 interface IERC2981Internal {
     error ERC2981Internal__RoyaltyExceedsMax();
-    error ERC2981Internal__RoyaltyNotSet();
 }
--- a/contracts/interfaces/IERC2981Internal.sol
+++ b/contracts/interfaces/IERC2981Internal.sol
@@ -6,5 +6,5 @@ pragma solidity ^0.8.8;
  * @title ERC2981 interface
  */
 interface IERC2981Internal {
-    error ERC2981Internal__RoyaltyExceedsMax();
+
 }
--- a/contracts/token/common/ERC2981/ERC2981.sol
+++ b/contracts/token/common/ERC2981/ERC2981.sol
@@ -1,0 +1,25 @@
+// SPDX-License-Identifier: MIT
+
+pragma solidity ^0.8.8;
+
+import { IERC2981 } from '../../../interfaces/IERC2981.sol';
+
+import { ERC2981Storage } from './ERC2981Storage.sol';
+import { ERC2981Internal } from './ERC2981Internal.sol';
+
+/**
+ * @title ERC2981 implementation
+ */
+abstract contract ERC2981 is IERC2981, ERC2981Internal {
+    /**
+     * @notice inheritdoc IERC2981
+     */
+    function royaltyInfo(uint256 tokenId, uint256 salePrice)
+        public
+        view
+        virtual
+        returns (address, uint256)
+    {
+        return _royaltyInfo(tokenId, salePrice);
+    }
+}
--- a/contracts/token/common/ERC2981/ERC2981.sol
+++ b/contracts/token/common/ERC2981/ERC2981.sol
@@ -15,9 +15,8 @@ abstract contract ERC2981 is IERC2981, ERC2981Internal {
      * @notice inheritdoc IERC2981
      */
     function royaltyInfo(uint256 tokenId, uint256 salePrice)
-        public
+        external
         view
-        virtual
         returns (address, uint256)
     {
         return _royaltyInfo(tokenId, salePrice);
--- a/contracts/token/common/ERC2981/ERC2981Internal.sol
+++ b/contracts/token/common/ERC2981/ERC2981Internal.sol
@@ -9,8 +9,6 @@ import { IERC2981Internal } from '../../../interfaces/IERC2981Internal.sol';
  * @title ERC2981 internal functions
  */
 abstract contract ERC2981Internal is IERC2981Internal {
-    using ERC2981Storage for ERC2981Storage.Layout;
-
     function _royaltyInfo(uint256 tokenId, uint256 salePrice)
         internal
         view
--- a/contracts/token/common/ERC2981/ERC2981Internal.sol
+++ b/contracts/token/common/ERC2981/ERC2981Internal.sol
@@ -24,24 +24,27 @@ abstract contract ERC2981Internal is IERC2981Internal {
         internal
         view
         virtual
-        returns (uint256)
+        returns (uint16 royaltyBPS)
     {
         ERC2981Storage.Layout storage l = ERC2981Storage.layout();
-        uint16 localRoyalty = l.royaltiesBPS[tokenId];
-        return localRoyalty > 0 ? localRoyalty : l.defaultRoyaltyBPS;
+        royaltyBPS = l.royaltiesBPS[tokenId];
+
+        if (royaltyBPS == 0) {
+            royaltyBPS = l.defaultRoyaltyBPS;
+        }
     }
 
     function _getRoyaltyReceiver(uint256 tokenId)
         internal
         view
         virtual
-        returns (address)
+        returns (address royaltyReceiver)
     {
         ERC2981Storage.Layout storage l = ERC2981Storage.layout();
-        address localReceiver = l.royaltyReceivers[tokenId];
-        return
-            localReceiver == address(0)
-                ? l.defaultRoyaltyReceiver
-                : localReceiver;
+        royaltyReceiver = l.royaltyReceivers[tokenId];
+
+        if (royaltyReceiver == address(0)) {
+            royaltyReceiver = l.defaultRoyaltyReceiver;
+        }
     }
 }
--- a/contracts/token/common/ERC2981/ERC2981Internal.sol
+++ b/contracts/token/common/ERC2981/ERC2981Internal.sol
@@ -9,17 +9,30 @@ import { IERC2981Internal } from '../../../interfaces/IERC2981Internal.sol';
  * @title ERC2981 internal functions
  */
 abstract contract ERC2981Internal is IERC2981Internal {
+    /**
+     * @notice called with the sale price to determine how much royalty is owed and to whom
+     * @param tokenId the ERC721 or ERC1155 token id to query for royalty information
+     * @param salePrice the sale price of the given asset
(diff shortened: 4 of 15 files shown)


contracts/token/common/ERC2981/ERC2981Internal.sol
```diff
@@ -0,0 +1,40 @@
...
+        }
+
+        if (royalty > 10000) {
+            revert ERC2981Internal__RoyaltyExceedsMax();
+        }
+
+        uint256 royaltyAmount = salePrice > 0
+            ? (royalty * salePrice) / 10000
+            : 0;
+
```
0xCourtney, line 37: I left this block without an unchecked directive, if the block overflows, it will return the overflowed value to the caller. Is this behavior acceptable?

contracts/token/common/ERC2981/ERC2981Mock.sol
```diff
@@ -0,0 +1,29 @@
...
+    constructor(
+        uint16 royalty,
+        uint16[] memory royalties,
+        address receiver
+    ) {
+        ERC2981Storage.Layout storage l = ERC2981Storage.layout();
+
+        l.royalty = royalty;
+        l.receiver = receiver;
+
```
0xCourtney, line 20: How should storage variables be set? Should each of them have a setter function like `ERC20MetadataStorage` or should they be accessed directly?

contracts/token/common/ERC2981/IERC2981Internal.sol
```diff
@@ -0,0 +1,13 @@
+// SPDX-License-Identifier: MIT
+
+pragma solidity ^0.8.8;
+
+import { IERC2981 } from '../../../interfaces/IERC2981.sol';
+
+/**
+ * @title ERC2981 interface
+ */
+interface IERC2981Internal {
```
ItsNickBarry, line 10: Move this to `interfaces/` and inherit it into `IERC2981`.  All interfaces that don't have "modifiers" such as "Base" or the "SolidState" prefix go in `interfaces/`.

contracts/token/common/ERC2981/IERC2981Internal.sol
```diff
@@ -0,0 +1,13 @@
+// SPDX-License-Identifier: MIT
+
+pragma solidity ^0.8.8;
+
+import { IERC2981 } from '../../../interfaces/IERC2981.sol';
```
ItsNickBarry, line 5: unused import

contracts/token/common/ERC2981/ERC2981Mock.sol
ItsNickBarry: Access directly.  We are in the process of removing all functions from the storage libraries and replacing them with functions in the internal contracts.

contracts/interfaces/IERC2981.sol
```diff
@@ -0,0 +1,23 @@
...
+     * @notice called with the sale price to determine how much royalty is owed and to whom.
+     * @param tokenId the NFT asset queried for royalty information
+     * @param salePrice the sale price of the NFT asset specified by tokenId
+     * @return address of who should be sent the royalty payment
+     * @return the royalty payment amount for salePrice
+     */
+    function royaltyInfo(uint256 tokenId, uint256 salePrice)
+        external
+        view
+        returns (address, uint256);
```
ItsNickBarry, line 22: Add names for the return parameters.

contracts/token/common/ERC2981/ERC2981.sol
```diff
@@ -0,0 +1,25 @@
...
+ * @title ERC2981 implementation
+ */
+abstract contract ERC2981 is IERC2981, ERC2981Internal {
+    /**
+     * @notice inheritdoc IERC2981
+     */
+    function royaltyInfo(uint256 tokenId, uint256 salePrice)
+        public
+        view
+        virtual
```
ItsNickBarry, line 20: change to `external`, remove `virtual`

ItsNickBarry (changes requested):

contracts/token/common/ERC2981/ERC2981Internal.sol
```diff
@@ -0,0 +1,40 @@
...
+        internal
+        view
+        virtual
+        returns (address, uint256)
+    {
+        ERC2981Storage.Layout storage l = ERC2981Storage.layout();
+
+        uint256 royalty = l.royalties[tokenId] > 0
+            ? l.royalties[tokenId]
+            : l.royalty;
```
ItsNickBarry, line 24: Move this to an internal `_getRoyaltyBPS(uint256 tokenId)` function.

ItsNickBarry: Makes it easier for the user to override this logic.

contracts/token/common/ERC2981/ERC2981Internal.sol
```diff
@@ -0,0 +1,40 @@
...
+
+        if (0 == royalty) {
+            revert ERC2981Internal__RoyaltyNotSet();
+        }
```
ItsNickBarry, line 28: This seems like a valid case, especially in a contract where royalties are set only on a per-token basis.  Is this revert case mentioned in the EIP?

contracts/token/common/ERC2981/ERC2981Internal.sol
ItsNickBarry: Leave the default behavior (no `unchecked` block), but remove the ternary operation.

contracts/token/common/ERC2981/ERC2981Internal.sol
0xCourtney: What does BPS stand for? 

Also, does it make sense to add the `Layout` as a parameter? It reduces the number of times the storage layout is accessed.

```js
function _getRoyaltyBPS(ERC2981Storage.Layout storage l, uint256 tokenId)
```

contracts/token/common/ERC2981/ERC2981Internal.sol
0xCourtney: The revert case is not mentioned in the EIP. Now that you point it out, it probably makes more sense to allow the royalty to be 0 rather than requiring the global or local royalty to be set. If that's the case, it may be necessary to use either the global or local royalty. This can be instrumented by using a flag, if its true use global, false, use local.

```js
function _getRoyaltyBPS(ERC2981Storage.Layout storage l, uint256 tokenId)
    internal
    view
    virtual
    returns (uint256)
{
    return l.DEFAULT_GLOBAL ? l.royalty : l.royalties[tokenId];
}
```

contracts/token/common/ERC2981/ERC2981Internal.sol
ItsNickBarry: https://www.investopedia.com/terms/b/basispoint.asp

contracts/token/common/ERC2981/ERC2981Internal.sol
ItsNickBarry: No, it's fine just like you have it:
```solidity
uint256 royalty = l.royalties[tokenId] > 0
            ? l.royalties[tokenId]
            : l.royalty;
```
Only issue there is that the mapping is queried twice - better to store `localRoyalty` or something in a local variable and then check if it's 0.

ItsNickBarry: As for the revert case, let's just delete it.

contracts/token/common/ERC2981/ERC2981Internal.sol
```diff
@@ -19,22 +19,19 @@ abstract contract ERC2981Internal is IERC2981Internal {
...
-        }
-
-        uint256 royaltyAmount = salePrice > 0
-            ? (royalty * salePrice) / 10000
-            : 0;
+        return (l.receiver, (royalty * salePrice) / 10000);
+    }
 
-        return (l.receiver, royaltyAmount);
+    function _getRoyaltyBPS(ERC2981Storage.Layout storage l, uint256 tokenId)
```
ItsNickBarry, line 28: Don't pass the storage layout as an argument here.  We did this extensively in the Premia contracts, and I think the added value there is questionable.  It may have been worth it at the time due to contract size savings.  In SolidState, however, we need to place a greater emphasis on clean interfaces, even when it increases gas usage slightly.

ItsNickBarry (changes requested):

contracts/interfaces/IERC2981Internal.sol
```diff
@@ -0,0 +1,10 @@
+// SPDX-License-Identifier: MIT
+
+pragma solidity ^0.8.8;
+
+/**
+ * @title ERC2981 interface
+ */
+interface IERC2981Internal {
+    error ERC2981Internal__RoyaltyExceedsMax();
```
ItsNickBarry, line 9: Errors should not be defined in the EIP interfaces in `interfaces/` directory.  Rather, they should be defined in an internal implementation interface, such as `IERC20BaseInternal`.

However, these contracts don't have a "base" implementation, so the pattern breaks down.  Closest analogue might be `ERC165`, which also includes a custom error but does not have a "base" implementation.  I'll think about what the best solution to this issue might be.

No action needed for now.

contracts/interfaces/IERC2981.sol
```diff
@@ -0,0 +1,24 @@
+// SPDX-License-Identifier: MIT
+
+pragma solidity ^0.8.8;
+
+import { IERC2981Internal } from './IERC2981Internal.sol';
+import { IERC165 } from './IERC165.sol';
```
ItsNickBarry, line 6:
```suggestion
import { IERC165 } from './IERC165.sol';
import { IERC2981Internal } from './IERC2981Internal.sol';
```

contracts/interfaces/IERC2981.sol
```diff
@@ -0,0 +1,24 @@
...
+ * @title ERC2981 interface
+ * @dev see https://eips.ethereum.org/EIPS/eip-2981
+ */
+interface IERC2981 is IERC2981Internal, IERC165 {
+    /**
+     * @notice called with the sale price to determine how much royalty is owed and to whom.
+     * @param tokenId the NFT asset queried for royalty information
+     * @param salePrice the sale price of the NFT asset specified by tokenId
+     * @return receiever address of who should be sent the royalty payment
+     * @return royaltyAmount the royalty payment amount for salePrice
```
ItsNickBarry, line 18:
```suggestion
     * @notice called with the sale price to determine how much royalty is owed and to whom
     * @param tokenId the ERC721 or ERC1155 token id to query for royalty information
     * @param salePrice the sale price of the given asset
     * @return receiever rightful recipient of royalty
     * @return royaltyAmount amount of royalty owed
```

contracts/token/common/ERC2981/ERC2981Internal.sol
```diff
@@ -0,0 +1,40 @@
...
+import { ERC2981Storage } from './ERC2981Storage.sol';
+import { IERC2981Internal } from '../../../interfaces/IERC2981Internal.sol';
+
+/**
+ * @title ERC2981 internal functions
+ */
+abstract contract ERC2981Internal is IERC2981Internal {
+    using ERC2981Storage for ERC2981Storage.Layout;
+
+    function _royaltyInfo(uint256 tokenId, uint256 salePrice)
```
ItsNickBarry, line 14:
```suggestion
abstract contract ERC2981Internal is IERC2981Internal {
    function _royaltyInfo(uint256 tokenId, uint256 salePrice)
```

contracts/token/common/ERC2981/ERC2981Mock.sol
```diff
@@ -0,0 +1,29 @@
...
+        for (uint8 i = 0; i < royalties.length; i++) {
+            l.royalties[i] = royalties[i];
+        }
+    }
+
+    function setRoyalty(uint16 royalty) external {
+        ERC2981Storage.layout().royalty = royalty;
```
ItsNickBarry, line 27: Should add an internal `_setRoyalty` function for this purpose.  Doesn't matter now, but might in the future if we decide to change how mocks work (see #92).

No action needed for now.

contracts/token/common/ERC2981/ERC2981Storage.sol
```diff
@@ -0,0 +1,24 @@
+// SPDX-License-Identifier: MIT
+
+pragma solidity ^0.8.8;
+
+library ERC2981Storage {
+    struct Layout {
+        // any value between 0 - 10000 should be accepted
+        uint16 royalty;
+        // 1-indexed to allow 0 to signify default royalty
```
ItsNickBarry, line 9: This comment seems to be unimplemented.  Interesting idea.  Though, it's cheaper to just use the `royalty` value for this purpose.
```suggestion
```

contracts/token/common/ERC2981/ERC2981Storage.sol
```diff
@@ -0,0 +1,24 @@
+// SPDX-License-Identifier: MIT
+
+pragma solidity ^0.8.8;
+
+library ERC2981Storage {
+    struct Layout {
+        // any value between 0 - 10000 should be accepted
+        uint16 royalty;
```
ItsNickBarry, line 8: Add BPS suffix and remove comment.

ItsNickBarry (approved):
@0xCourtney I made a few small changes, but okay to merge.

Two optional additional tasks:
1. Inherit the ERC2918 implementation into `SolidStateERC721` and `SolidStateERC1155`.  I think it's good to promote it as a sensible default.
2. Enable per-token-id receiver settings.  Same pattern as royalty amounts: a mapping for granular settings, and a fallback value.  Update the `_royaltyReceiver` function to take a `tokenId` argument.

Edit: implemented task number 2

ItsNickBarry (commented):
@0xCourtney Does the EIP say anything about reverting if royalty rate is greater than 10000?  I have two issues with that:
1. Perhaps it's valid to set a 200% royalty?  If an NFT is sold for 100, the artist gets 200.  Ridiculous, but valid.
2. This could cause external contracts to revert if a royalty us updated to a value greater than 10000, potentially locking assets.  If the revert is necessary, it should be done in the setter, not the getter (there is no setter, but I think we should add one as mentioned in a previous comment).

ItsNickBarry:
Does the EIP even indicate whether the royalty is _deducted from_ or _in addition to_ the `salePrice`?  My above comment assumes the latter.

spec/token/common/ERC2981.behavior.ts
```diff
@@ -0,0 +1,67 @@
...
+    let tokenIdThree = BigNumber.from(3);
+
+    let receiver: SignerWithAddress;
+    let instance: ERC2981Mock;
+
+    before(async function () {
+      receiver = (await ethers.getSigners())[1];
+      instance = await deploy();
+    });
+
```
ItsNickBarry, line 26: Must call `describeBehaviorOfERC165` with the relevant `interfaceId` for the EIP.  See `ERC721Base.behavior.ts` for an example.

ItsNickBarry (changes requested):

0xCourtney:
> @0xCourtney Does the EIP say anything about reverting if royalty rate is greater than 10000? I have two issues with that:

The specification did not suggest throwing an error if the rate exceeds 10000. I tend to agree, the revert should be removed or handled in a setter independent of `_royaltyInfo`.

spec/token/common/ERC2981.behavior.ts
0xCourtney: https://github.com/solidstate-network/solidstate-solidity/pull/147/commits/aa82c30bcbc867c17580076ba267737c027aecc5
https://github.com/solidstate-network/solidstate-solidity/pull/147/commits/e1c0b3ce994f0cca4d46c4c349c335d7d7bc57e9

ItsNickBarry:
> I think the standard is purposefully lightweight.

Not good enough.  The standard is barely even a standard if it doesn't answer the fundamental question of whether the royalty is added to or deducted from the base cost.

But, nothing we can do about that.  Let's just assume that the royalty is added, and therefore allow royalties above 10000 BPS.  I will make this change now.

ItsNickBarry (approved):
@0xCourtney Go ahead and merge if you think my changes are okay.

0xCourtney:
> Not good enough. The standard is barely even a standard if it doesn't answer the fundamental question of whether the royalty is added to or deducted from the base cost.

I'm not crazy about it either, this may be more relevant than the original quote I passed along.

> Marketplaces that support this standard SHOULD implement some method of transferring royalties to the royalty recipient. Standards for the actual transfer and notification of funds will be specified in future EIPs.

Out of curiosity, how would the implementation be impacted whether the royalty were added or removed from the sale price?

ItsNickBarry:
I believe that those excerpts from the EIP are saying that the manner of payment is not specified in the EIP.  The following would be acceptable:
* Immediate token transfer.
* Immediate conversion to some other token and transfer.
* Tracking of credits for later transfer.
* Etc.

But I'm asking a different question.  Say the royalty is set o 20%, and an NFT is sold for $100.  There are several ways to handle this:
1. Send 100 to the seller and 20 to the royalty receiver (royalty is added on top of the base cost).
2. Send 80 to the seller and 20 to the royalty receiver (royalty is deducted from the base cost).
3. Send 100 to the seller and 25 to the royalty receiver (royalty is a percentage of total amount transferred, and the seller still gets the full base cost).

Listed in order of likelihood.  Our implementation assumes the first.

Now say the royalty is set to 200%.
1. Send 100 to the seller and 200 to the royalty receiver.
2. Send -100 to the seller, and 200 to the royalty receiver (calculation is broken).
3. Calculation is impossible.

The first option is ultimately the most intuitive, most flexible, and easiest to code, so I think most people will assume that this is the behavior intended by the EIP.  But the EIP should be more clear, because the other two options are perfectly valid in certain circumstances.

0xCourtney:
> The first option is ultimately the most intuitive, most flexible, and easiest to code, so I think most people will assume that this is the behavior intended by the EIP. But the EIP should be more clear, because the other two options are perfectly valid in certain circumstances.

Thanks for explaining that. It's clear that the EIP left a lot of room for interpretation and potentially room for bugs to arise. Given your example would it make sense to add a setter for the royalty BPS, and throw if 10000 is exceeded as suggested earlier? 

We can note the case you mentioned so that the dev is aware of the risk of overriding the default function and permitting the royalty to exceed 100%.

ItsNickBarry:
Because we're assuming the first case, we don't need to include any reverts.  I added a note in b0355c3c0571bd6caea7e60dfa1556d8ea46cf15.

PR #151, opened 2022-10-20
Update and test `EnumerableMap`

NouDaimon:
Contains:

- tests for EnumerableMap
- fix for `_at(AddressToAddressMap storage map, uint256 index)` function

Initially only for `AddressToAddressMap`, once passed initial scrutiny, the `UintToAddressMap` tests will be added.

### DIFF
--- a/contracts/data/EnumerableMap.sol
+++ b/contracts/data/EnumerableMap.sol
@@ -139,6 +139,44 @@ library EnumerableMap {
         return _remove(map._inner, bytes32(key));
     }
 
+    function toArray(AddressToAddressMap storage map)
+        internal
+        view
+        returns (address[] memory keys, address[] memory values)
+    {
+        uint256 len = map._inner._entries.length;
+        keys = new address[](len);
+        values = new address[](len);
+        unchecked {
+            for (uint256 i; i < len; ++i) {
+                keys[i] = address(
+                    uint160(uint256(map._inner._entries[i]._key))
+                );
+                values[i] = address(
+                    uint160(uint256(map._inner._entries[i]._value))
+                );
+            }
+        }
+    }
+
+    function toArray(UintToAddressMap storage map)
+        internal
+        view
+        returns (uint256[] memory keys, address[] memory values)
+    {
+        uint256 len = map._inner._entries.length;
+        keys = new uint256[](len);
+        values = new address[](len);
+        unchecked {
+            for (uint256 i; i < len; ++i) {
+                keys[i] = uint256(map._inner._entries[i]._key);
+                values[i] = address(
+                    uint160(uint256(map._inner._entries[i]._value))
+                );
+            }
+        }
+    }
+
     function _at(Map storage map, uint256 index)
         private
         view
--- a/contracts/data/EnumerableMap.sol
+++ b/contracts/data/EnumerableMap.sol
@@ -36,18 +36,10 @@ library EnumerableMap {
     {
         (bytes32 key, bytes32 value) = _at(map._inner, index);
 
-        address addressKey;
-        address addressValue;
-
-        assembly {
-            addressKey := and(0xFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFF, key)
-            addressValue := and(
-                0xFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFF,
-                value
-            )
-        }
-
-        return (addressKey, addressValue);
+        return (
+            address(uint160(uint256(key))),
+            address(uint160(uint256(value)))
+        );
     }
 
     function at(UintToAddressMap storage map, uint256 index)
--- a/contracts/data/EnumerableMap.sol
+++ b/contracts/data/EnumerableMap.sol
@@ -35,11 +35,19 @@ library EnumerableMap {
         returns (address, address)
     {
         (bytes32 key, bytes32 value) = _at(map._inner, index);
+
         address addressKey;
+        address addressValue;
+
         assembly {
-            addressKey := mload(add(key, 20))
+            addressKey := and(0xFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFF, key)
+            addressValue := and(
+                0xFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFF,
+                value
+            )
         }
-        return (addressKey, address(uint160(uint256(value))));
+
+        return (addressKey, addressValue);
     }
 
     function at(UintToAddressMap storage map, uint256 index)
--- a/contracts/data/EnumerableMap.sol
+++ b/contracts/data/EnumerableMap.sol
@@ -177,6 +177,70 @@ library EnumerableMap {
         }
     }
 
+    function keys(AddressToAddressMap storage map)
+        internal
+        view
+        returns (address[] memory keys)
+    {
+        uint256 len = map._inner._entries.length;
+        keys = new address[](len);
+
+        unchecked {
+            for (uint256 i; i < len; ++i) {
+                keys[i] = address(
+                    uint160(uint256(map._inner._entries[i]._key))
+                );
+            }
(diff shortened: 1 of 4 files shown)


contracts/data/EnumerableMap.sol
```diff
@@ -35,11 +35,19 @@ library EnumerableMap {
         returns (address, address)
     {
         (bytes32 key, bytes32 value) = _at(map._inner, index);
+
         address addressKey;
+        address addressValue;
+
         assembly {
-            addressKey := mload(add(key, 20))
+            addressKey := and(0xFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFF, key)
```
NouDaimon, line 43: This is not an elegant solution. 

I'll have a look at what else can be done.

ItsNickBarry: I don't know why there was an assembly block here.  All we need is a simple cast: `return (address(uint160(uint256(key))), address(uint160(uint256(value))));`

test/data/EnumerableMap.ts
```diff
@@ -0,0 +1,415 @@
...
+      const addressThree = bnToAddress(BigNumber.from('300'));
+      const addressFour = bnToAddress(BigNumber.from('400'));
+      const addressFive = bnToAddress(BigNumber.from('500'));
+      const addressSix = bnToAddress(BigNumber.from('600'));
+
+      describe('#at(uint256)', () => {
+        it('returns value coresponding to index provided', async () => {
+          await instance['set(address,address)'](addressOne, addressFour);
+          await instance['set(address,address)'](addressTwo, addressFive);
+          await instance['set(address,address)'](addressThree, addressSix);
```
ItsNickBarry, line 37: In several places you have multiple similar calls.  I don't think it adds much value.  One is enough.

test/data/EnumerableMap.ts
```diff
@@ -0,0 +1,415 @@
...
+
+          expect(await instance['at(uint256)'](0)).to.deep.equal([
+            addressOne,
+            addressFour,
+          ]);
```
ItsNickBarry, line 42: I don't like the use of `deep.equal` here.  Better to destructure the return data and compare each piece.
```js
const [key, value] = await instance.callStatic['at(uint256)'](0);
expect(key).to.equal(...);
```

ItsNickBarry: Also, use `callStatic` when calling view and pure functions.

test/data/EnumerableMap.ts
```diff
@@ -0,0 +1,415 @@
...
+          expect(await instance['length()']()).to.equal(3);
+
+          await instance['remove(address)'](addressThree);
+          expect(await instance['length()']()).to.equal(2);
+
+          await instance['remove(address)'](addressTwo);
+          expect(await instance['length()']()).to.equal(1);
+
+          await instance['remove(address)'](addressOne);
+          expect(await instance['length()']()).to.equal(0);
```
ItsNickBarry, line 101: This is a good case for making multiple calls.  I would probably put them in a loop, though doing that is dangerous because if a loop has 0 iterations, the issue won't show up as a test failure.

test/data/EnumerableMap.ts
```diff
@@ -0,0 +1,415 @@
...
+        it('returns length of enumerable map', async () => {
+          expect(await instance['length()']()).to.equal(0);
+
+          await instance['set(address,address)'](addressOne, addressFour);
+          expect(await instance['length()']()).to.equal(1);
+
+          await instance['set(address,address)'](addressTwo, addressFive);
+          expect(await instance['length()']()).to.equal(2);
+
+          await instance['set(address,address)'](addressThree, addressSix);
```
0xCourtney, line 91: perhaps you can add a case or logic that shows adding an entry of an existing key does not increase the length

```js
...
await instance['set(address,address)'](addressThree, addressSix);
expect(await instance['length()']()).to.equal(3);

await instance['set(address,address)'](addressThree, addressSeven);
expect(await instance['length()']()).to.equal(3);
...
```

0xCourtney (commented):
Good job getting this going! Does it make sense to add [`toArray`](https://github.com/solidstate-network/solidstate-solidity/blob/EnumerableMapTests/contracts/data/EnumerableSet.sol#L153-L168) as we did for `EnumerableSet`?

ItsNickBarry:
@0xCourtney It's not immediately clear what `toArray` would return.  Is it possible to return an array of tuples?  If not, maybe two separate functions for keys and values?

0xCourtney:
> @0xCourtney It's not immediately clear what `toArray` would return. Is it possible to return an array of tuples or array of arrays? If not, maybe two separate functions for keys and values?

Great question, I think supporting `toArray` which returns an array of tuples makes sense as it keeps the interface consistent with the other data structures.

```solidity
struct AddressToAddress {
    address key;
    address value;
}

function toArray(AddressToAddressMap storage map)
    internal
    view
    returns (AddressToAddress[] memory)
{
    uint256 len = _length(map._inner);
    AddressToAddress[] memory arr = new AddressToAddress[](len);

    unchecked {
        for (uint256 index; index < len; ++index) {
            (address key, address value) = at(map, index);
            arr[index] = AddressToAddress({ key: key, value: value });
        }
    }

    return arr;
}

// [
//   [
//     '0x0000000000000000000000000000000000000064',
//     '0x000000000000000000000000000000000000012c',
//     key: '0x0000000000000000000000000000000000000064',
//     value: '0x000000000000000000000000000000000000012c'
//   ],
//   [
//     '0x00000000000000000000000000000000000000C8',
//     '0x000000000000000000000000000000000000012c',
//     key: '0x00000000000000000000000000000000000000C8',
//     value: '0x000000000000000000000000000000000000012c'
//   ],
//   [
//     '0x000000000000000000000000000000000000012c',
//     '0x00000000000000000000000000000000000000C8',
//     key: '0x000000000000000000000000000000000000012c',
//     value: '0x00000000000000000000000000000000000000C8'
//   ]
// ]
```


Adding `keys` and `values` (not particular on the naming) which return an array of keys and an array of values, respectively, would probably be more useful for the dev.

test/data/EnumerableMap.ts
```diff
@@ -0,0 +1,415 @@
...
+              instance['get(address)'](addressOne),
+            ).to.be.revertedWithCustomError(
+              instance,
+              'EnumerableMap__NonExistentKey',
+            );
+          });
+        });
+      });
+
+      describe('#set(address,address)', () => {
```
0xCourtney, line 130: We should include a test which demonstrates that calling `set` for a key that already exists overwrites the value.

```js
it('overwrites the address value at address key', async () => {
  await instance['set(address,address)'](addressOne, addressTwo);
  await instance['set(address,address)'](addressOne, addressThree);

  expect(await instance['get(address)'](addressOne)).to.eq(
    addressThree,
  );
});
```

0xCourtney (changes requested):

ItsNickBarry:
@0xCourtney Your example returns a struct.  Do you know if Solidity supports tuple arrays?  If it doesn't, I'd favor returning two separate arrays over defining a single-use struct.

0xCourtney:
> @0xCourtney Your example returns a struct. Do you know if Solidity supports tuple arrays? If it doesn't, I'd favor returning two separate arrays over defining a single-use struct.

No, tuples are not a proper type and are only accessible to Solidity internally according to [this](https://docs.soliditylang.org/en/latest/control-structures.html#destructuring-assignments-and-returning-multiple-values). Likewise, if `toArray` returns a tuple of arrays it would be much cleaner.

ItsNickBarry:
Thanks for the confirmation.  In that case, we'll do it like in my example interface above.

ItsNickBarry (approved):
Didn't check whether all tests are copy/pasted properly, but coverage is at 100%.

No real reason to use `EnumerableSetTests` as the base branch.  Now we need to wait until that branch is merged before merging this one.

0xCourtney (approved):

Issue #153, opened 2022-10-23
Partial Pausability


wiasliaw:
# Motivation

## a.

`PausableStorage` uses a full slot but only use 1 byte for boolean.

```solidity
struct Layout {
    bool paused;
}
```

## b.

When it comes to diamond's facets, partial pausability might in need.

For example,
- facet A is not associated with facet B, pause A should not pause B.
- facet B is associated with facet C, pause B should also pause C.

# Rationale

`PausableStorage` should be a uint256 which represent a bitmap `mapping(uint8 => bool)`. Each facet can pick a value(uint8) for pausability and not to bother with other facets.

```solidity
struct Layout {
    uint256 paused;
}
```

# Example

- pause A but not pause B
- pause B also pause C
- combining A and BC, pause D also pause A, B, C

```solidity
contract FacetA {
    uint256 private constant mask = 1 << 0; // bitmap[0]
    function foo1() whenNotPause(mask) { ... }
}

contract FacetB {
    uint256 private constant mask = 1 << 1; // bitmap[1]
    function foo2() whenNotPause(mask) { ... }
}

contract FacetC {
    uint256 private constant mask = 1 << 1; // bitmap[1]
    function foo3() whenNotPause(mask) { ... }
}

contract FacetD {
    uint256 private constant mask = 1 << 1 || 1 << 0; // bitmap[1] || bitmap[0]
    function foo4() whenNotPause(mask) { ... }
}
```

ItsNickBarry:
This is a good idea.  As long as we keep an interface for pausing without supplying a parameter, then I would be in favor of merging this.

Maybe slot 255 could be used as a default.

wiasliaw:
How about define an unimplement function to custom which bit should be use?

proposal here: https://github.com/wiasliaw/solidstate-pausable

ItsNickBarry:
That wouldn't work with the "visibility layers" pattern.  The recommended way to share internal logic between facets is to share a single internal contract between multiple facets.  Using different versions of the same internal code would be confusing.  Your previous idea described above would work better.

wiasliaw:
The second idea is to meet "keep an interface for pausing without supplying a parameter". I don't think it should keep the interface always the same. Additionally, the interface for external function hasn't altered, only alter the internal functions.

i.e.
```solidity
contract FacetA is Pausable {
    uint256 private constant mask = 1 << 0; // bitmap[0]

    function foo1() whenNotPause(mask) { ... }

[...]

PR #157, opened 2022-10-28
Added PartialPausable #feature-153

iam-dev:


ItsNickBarry (commented):
A few issues:
* Some important files in `abi/` have been removed.
* `package-lock.json` files have been added - SolidState uses yarn.
* The partial pausability feature should be added to `Pausable`, not to a new set of files.

I'm also not sure about using and `address` value for the modifiers.  I see why you did it, but in general a facet's address should only be used in the diamond itself, for the purpose of executing `delegatecall`.  I'll post more comments in #153.

ItsNickBarry:
A different implementation of `PartiallyPausable` has been merged in #199 and released in `v0.0.55`.  Thanks for the contribution @iam-dev.

PR #158, opened 2022-10-30
Diamond Helper Functions

### LINKED ISSUE #65: `diamondCut` utilities
The `diamondCut` function is complex, and many users would benefit from a package to assist with adding/replacing/removing function selectors from diamond proxies.

0xCourtney:
# Summary
EIP2535 is a standard quickly growing in popularity and adoption. While Diamonds offer a great pathway for scaling contracts there are not many tools for managing deployments or upgrades. This PR introduces utility functions that provide developers with a consistent and reliable way of maintaining their contracts. The key goal here is to offer tools that perform the minimal changes necessary to add, replace, or remove facets from a Diamond.

### General

- `diamondCut` - calls `Diamond.diamondCut`
- `printDiamond` - prints a table containing the registered and unregistered selectors

### Deployment

- `addUnregisteredSelectors` - adds unregistered selectors to the Diamond.

### Upgrade

- `replaceRegisteredSelectors` - replaces any selector which is registered to a different target address
- `removeRegisteredSelectors` - removes the registered selector from the Diamond

### Examples

```js
// TestFacets.sol

contract Test1Facet {
    function test1Func1(bool) public returns (bytes32[] memory) {}
    function test1Func2(int128) external view returns (address) {}
    function test1Func3(address, address, address) external {}
}

contract Test1Facet1 {
    function test1Func1(bool) public returns (bytes32[] memory) {}
    function test1Func2() external view returns (address) {}
}

contract Test2Facet {
    function test2Func1() external pure returns (uint8) {}
    function test2Func2(bytes8) external view returns (bool) {}
}
```

```js
// deploy.ts

// deploy Diamond
// registers selectors from Test1Facet and Test2Facet on Diamond
let facetCut = await addUnregisteredSelectors(Diamond, [Test1Facet, Test2Facet]);
await diamondCut(Diamond, facetCut);

// upgrade Diamond
// adds Test1Facet1.test1Func2, since the signature has changed
facetCut = await addUnregisteredSelectors(Diamond, [Test1Facet1]);
await diamondCut(Diamond, facetCut);

// replace Test1Facet.test1Func1 with Test1Facet1.test1Func1
facetCut = await replaceRegisteredSelectors(Diamond, [Test1Facet1]);
await diamondCut(Diamond, facetCut);

// removes Test1Facet.test1Func2 and Test1Facet.test1Func3
facetCut = await removeRegisteredSelectors(Diamond, [Test2Facet, Test1Facet1]);
await diamondCut(Diamond, facetCut);
```

## Actions
- [ ] Add tests to repo


lib/diamonds.ts
```diff
@@ -0,0 +1,98 @@
...
+  Add,
+  Replace,
+  Remove,
+}
+
+export interface FacetCut extends Facet {
+  action: FacetCutAction;
+}
+
+export function getContractSignatures(contract: Contract) {
```
NouDaimon, line 20: Should these functions have an explicit return type?

lib/diamonds.ts
```diff
@@ -0,0 +1,137 @@
...
+    selectors: selectors,
+  };
+}
+
+// adds unregistered selectors
+export async function addUnregisteredSelectors(
+  diamond: Contract,
+  contracts: Contract[],
+) {
+  const registeredFacets: Facet[] = await diamond.facets();
```
NouDaimon, line 59: Do we need any try catches?

Could we use `@solidstate/typechain-types`?

lib/diamonds.ts
```diff
@@ -0,0 +1,137 @@
...
+  let facets: Facet[] = [];
+  contracts.forEach((contract) => {
+    facets.push({
+      target: contract.address,
+      selectors: getContractSelectors(contract),
+    });
+  });
+
+  let facetCuts: FacetCut[] = [];
+  // if selector is not found in the registered selectors then it should be
```
NouDaimon, line 70: This function can be duplicated, and modified a bit in order to add unregistered selectors, and replace registered selectors. 

Perhaps called `upgradeExistingAndAddSelectors`?

NouDaimon (commented):
Awesome initiative! Eventually, these utility functions will prove very useful to third parties.

lib/diamonds.ts
0xCourtney: There's definitely room to build a single function that will handle deployments and upgrades. I do have some concerns about being opinionated about how deployments and upgrades are done. From my conversations with devs who use EIP2535, I found that there is a wide range of preferences for doing this. For example, some people will replace all of the selectors when upgrading others will replace only the ones that are needed. The approach I took here was to make the least amount of changes when adding/ replacing/ removing selectors.

0xCourtney:
@NouDaimon thanks for the feedback! I have a bunch of changes on my local which I plan on pushing this week. I will update the description with my checklist of things before this is ready to come out of draft.

lib/diamonds.ts
0xCourtney: good thinking, these should definitely have return types defined.

lib/diamonds.ts
0xCourtney: > Do we need any try catches?

Possibly, did you have anything in mind? I was considered adding some form of validation that would throw if some conditions could result in `diamondCut` reverting. In the end I decided to use if-statements to check these conditions before adding a facetCut. Not sure if this is what you had in mind though.

```js
if (
  target !== diamond.address &&
  selector.length > 0 &&
  !selectorExistsInFacet(selector, registeredFacets)
) {
  facetCuts.push(
    printFacetCuts(facet.target, [selector], FacetCutAction.Add),
  );
}
```

0xCourtney: > Could we use `@solidstate/typechain-types`?

Definitely, I have this on my list of things to do.

0xCourtney:
Any recommendations on where I should add tests and contracts?

cc @ItsNickBarry

lib/diamonds.ts
```diff
@@ -0,0 +1,327 @@
+import {
+  IDiamondReadable,
+  IDiamondWritable,
+} from '@solidstate/typechain-types';
+import { Contract, constants } from 'ethers';
+
+const Table = require('cli-table3');
```
ItsNickBarry, line 7: I'm conflicted about whether SolidState should include pretty print functions.  CLI features should be included in dedicated CLI modules, such as Hardhat plugins.  But such a plugin needs planning and might not cover all use cases for pretty printing.

lib/diamonds.ts
```diff
@@ -0,0 +1,327 @@
...
+
+type Diamond = IDiamondReadable & IDiamondWritable;
```
ItsNickBarry, line 9: Avoid the composite type and use only the minimum type where necessary.  For example, `addUnregisteredSelectors` doesn't need `IDiamondWritable`.

lib/diamonds.ts
```diff
@@ -0,0 +1,327 @@
...
+
+export interface Facet {
+  target: string;
+  selectors: string[];
+}
+
+enum FacetCutAction {
```
ItsNickBarry, line 16: Export this and use it in the test cases that need `FacetCutAction`.

lib/diamonds.ts
```diff
@@ -0,0 +1,327 @@
...
+            [selector],
+            FacetCutAction.Remove,
+          ),
+        );
+      }
+    }
+  }
+
+  return groupFacetCuts(facetCuts);
+}
```
ItsNickBarry, line 157: The add/replace/remove function names imply that a transaction is sent, but really they just generate the `FacetCut` formatting.  Should change the name to something like `generateFacetCutsForAdd`.  Can you think of something better?

ItsNickBarry: These functions are rather specific, so I think they should throw errors if certain passed data can't be added/removed/replaced.

ItsNickBarry: I'd also like a function that takes a list of selectors and targets, and calculates the `FacetCut` diff needed to migrate an existing diamond to the given schema.

ItsNickBarry (commented):
Still a bit unsure about how to organize this, but definitely a good start.

@0xCourtney Can you list all the places where you think a developer would use the print function?  As mentioned in an inline comment, in my opinion that function would fit best in a CLI package.  How to print something is fundamentally more opinionated than SS is trying to be.  But you might have some use cases in mind that would justify including it here.

I'll put some more thought in to a possible Hardhat plugin.

lib/diamonds.ts
0xCourtney: Admittedly I wrote this function as an experiment in visualizing the diamond facets. I agree though, this could very easily be a HH plugin.

ItsNickBarry: Let's keep it here for now but not export it as part of the API.  We'll move it to a plugin later, and for now focus on the other functions.

lib/diamonds.ts
0xCourtney: > The add/replace/remove function names imply that a transaction is sent, but really they just generate the `FacetCut` formatting. Should change the name to something like `generateFacetCutsForAdd`. Can you think of something better?

I asked ChatGPT for help here, how about `addUnregisteredSelectorsToFacetCut`?

lib/diamonds.ts
0xCourtney: https://github.com/solidstate-network/solidstate-solidity/pull/158/commits/047964efc5ae891831a15dd38d09cd2ca767c831

lib/diamonds.ts
0xCourtney: https://github.com/solidstate-network/solidstate-solidity/pull/158/commits/9d3b06061fdc6bc420fc00d11bec97ff7995c6b7

lib/diamonds.ts
0xCourtney: https://github.com/solidstate-network/solidstate-solidity/pull/158/commits/b2eda5732ea11020a502eea1572d8f5d472de046

lib/diamonds.ts
```diff
@@ -0,0 +1,327 @@
+import {
+  IDiamondReadable,
+  IDiamondWritable,
+} from '@solidstate/typechain-types';
+import { Contract, constants } from 'ethers';
```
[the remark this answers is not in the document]
0xCourtney: https://github.com/solidstate-network/solidstate-solidity/pull/158/commits/b2eda5732ea11020a502eea1572d8f5d472de046

lib/diamonds.ts
0xCourtney: https://github.com/solidstate-network/solidstate-solidity/pull/158/commits/f9a6c2c2c162b15dd1c97c4f5fd9b01eba971469

lib/diamonds.ts
0xCourtney: > I'd also like a function that takes a list of selectors and targets, and calculates the `FacetCut` diff needed to migrate an existing diamond to the given schema.

https://github.com/solidstate-network/solidstate-solidity/pull/158/commits/e702c36dad50b8cea06135110ab87c1d206496e4

0xCourtney: > These functions are rather specific, so I think they should throw errors if certain passed data can't be added/removed/replaced.

https://github.com/solidstate-network/solidstate-solidity/pull/158/commits/45b2dc44e30d47ece990016ecc65e33dfa97900d

0xCourtney:
In https://github.com/solidstate-network/solidstate-solidity/pull/158/commits/3d9d4568e952998dafc71baacaf4103b93bf1319 I introduced the concept of filters. A filter consists of an object containing a contract address and an array of selectors. 

```js
interface FacetFilter {
  contract: string;
  selectors: string[];
}
```

The `only` filter will add/replace/remove selectors defined in the `only` filter ignoring all other selectors for that contract. The `exclude` filter will ignore all selectors defined in the `exclude` filter for that contract, all other selectors will get added/replaced/removed.

```js
await d.addUnregisteredSelectors(Diamond, [Test1Facet1])

// [{
//   target: '0x9fE46736679d2D9a65F0992F2272dE9f3c7fa6e0',
//   action: 0,
//   selectors: [ '0xc2985578', '0x0716c2ae' ]
// }]

await d.addUnregisteredSelectors(
  Diamond,
  [Test1Facet1],
  [{ contract: Test1Facet1.address, selectors: ['0xc2985578'] }],
  [],
);

// [{
//   target: '0x9fE46736679d2D9a65F0992F2272dE9f3c7fa6e0',
//   action: 0,
//   selectors: [ '0xc2985578' ]
// }]

await d.addUnregisteredSelectors(
  Diamond,
  [Test1Facet1],
  [],
  [{ contract: Test1Facet1.address,selectors: ['0xc2985578'] }],
)

// [{
//   target: '0x9fE46736679d2D9a65F0992F2272dE9f3c7fa6e0',
//   action: 0,
//   selectors: [ '0x0716c2ae' ]
// }]
```

The `only` and `exclude` filters must not overlap, therefore a contract may define selectors in the `only` or `exclude` filter, not both.

lib/package.json
```diff
@@ -25,6 +25,8 @@
     "tsc-clean": "tsc --build --clean tsconfig.json"
   },
   "dependencies": {
+    "@ethersproject/constants": "^5.7.1",
+    "@ethersproject/contracts": "^5.7.1",
```
ItsNickBarry, line 29: These versions don't seem to exist:
https://www.npmjs.com/package/@ethersproject/contract?activeTab=versions
https://www.npmjs.com/package/@ethersproject/constants?activeTab=versions

ItsNickBarry: Not really sure how to interpret ethers versioning: https://www.npmjs.com/org/ethersproject

Looks like they're keeping major and feature versions in sync, but not patch versions.  A version `5.7.1` exists, but not for these particular packages.

lib/diamonds.ts
ItsNickBarry: See #176

lib/package.json
ItsNickBarry: Fixed in #176

lib/diamonds.ts
```diff
@@ -50,16 +50,71 @@ export function selectorExistsInFacets(
...
+}
+
+interface FacetFilter {
+  contract: string;
+  selectors: string[];
+}
+
+function selectorIsFiltered(
+  only: FacetFilter[],
+  exclude: FacetFilter[],
```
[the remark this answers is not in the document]
0xCourtney: https://github.com/solidstate-network/solidstate-solidity/pull/158/commits/ce286c65c33e693e847c0fdd3e77f34bde8a83ad

lib/diamond/utils.ts
```diff
@@ -0,0 +1,305 @@
...
+  return Object.keys(contract.interface.functions);
+}
+
+// returns a list of selectors for a contract
+export function getSelectors(contract: Contract): string[] {
+  const signatures = getSignatures(contract);
+  return signatures.reduce((acc: string[], val: string) => {
+    acc.push(contract.interface.getSighash(val));
+    return acc;
+  }, []);
```
ItsNickBarry, line 35:
```suggestion
  return getSignatures(contract).map((signature) => contract.interface.getSighash(signature));
```

lib/diamond/utils.ts
```diff
@@ -0,0 +1,305 @@
...
+  target: string;
+  selectors: string[];
+}
+
+export interface FacetCut extends Facet {
+  action: FacetCutAction;
+}
+
+// returns a list of signatures for a contract
+export function getSignatures(contract: Contract): string[] {
```
ItsNickBarry, line 25: Rename to `getFuntionSignatures` so as not to imply that event signatures, etc, are included.

lib/diamond/filters.ts
```diff
@@ -0,0 +1,57 @@
+import { AddressZero } from '@ethersproject/constants';
+
+export interface FacetFilter {
+  contract: string;
+  selectors: string[];
+}
```
ItsNickBarry, line 6: It looks like these `FacetFilter` objects are used in sets of 3 arrays, corresponding to the `FacetCutAction` types.  Would it not be better to include the action in the `FacetFilter` type?

lib/diamond/filters.ts
```diff
@@ -0,0 +1,57 @@
...
+  contract: string,
+  selector: string,
+): boolean {
+  for (const filter of filters) {
+    if (filter.contract === contract || AddressZero === contract) {
+      return filter.selectors.includes(selector);
+    }
+  }
+
+  return false;
```
ItsNickBarry, line 41:
```suggestion
  return filters.some((filter) => [filter.contract, AddressZero].includes(contract) && filter.selectors.includes(selector));
```

lib/diamond/utils.ts
```diff
@@ -0,0 +1,305 @@
...
+}
+
+// executes a DiamondCut using the provided FacetCut
+export async function diamondCut(
+  diamond: IDiamondWritable,
+  facetCut: FacetCut[],
+  target: string = AddressZero,
+  data: string = '0x',
+) {
+  (await diamond.diamondCut(facetCut, target, data)).wait(1);
```
ItsNickBarry, line 249: How exactly does `wait` work?

lib/diamond/utils.ts
```diff
@@ -0,0 +1,305 @@
...
+      } else {
+        cache[selector] = true;
+      }
+    });
+  });
+
+  return cuts;
+}
+
+export function printFacetCuts(
```
ItsNickBarry, line 295: This doesn't seem to print anything.

lib/diamond/utils.ts
```diff
@@ -0,0 +1,305 @@
...
+          printFacetCuts(target, [selector], FacetCutAction.REPLACE),
+        );
+
+        selectorsReplaced = true;
+      }
+    }
+  }
+
+  if (!selectorsReplaced) {
+    throw new Error('No selectors were replaced in FacetCut');
```
ItsNickBarry, line 141: The errors are thrown if _all_ changes are skipped, but I think we'll want to throw if _any_ change is skipped.  Need to think about this though.

0xCourtney: Would this be the result of using filters or something else?

lib/diamond/utils.ts
0xCourtney: > [Resolves to the TransactionReceipt once the transaction has been included in the chain for confirms blocks. If confirms is 0, and the transaction has not been mined, null is returned.](https://docs.ethers.org/v5/api/providers/types/#providers-TransactionResponse)

It may be good to return the TransactionReceipt here. Otherwise, a user may not be able to detect errors or access the receipt, if needed.

lib/diamond/utils.ts
0xCourtney: https://github.com/solidstate-network/solidstate-solidity/pull/158/commits/9d92ea0a7f5d5231717e8a2cd41de95fcb1f70c5

lib/diamond/utils.ts
0xCourtney: https://github.com/solidstate-network/solidstate-solidity/pull/158/commits/871f8686765295d69347c3971a6517ad8c35d20d

lib/diamond/utils.ts
0xCourtney: https://github.com/solidstate-network/solidstate-solidity/pull/158/commits/f4c211b7883346956d256edcaac04fa951251fdc

lib/diamond/filters.ts
0xCourtney: https://github.com/solidstate-network/solidstate-solidity/pull/158/commits/54da442a2170c070c2403b7f395faf573b038120

lib/diamond/utils.ts
```diff
@@ -22,13 +22,15 @@ export interface FacetCut extends Facet {
 }
 
 // returns a list of signatures for a contract
-export function getSignatures(contract: Contract): string[] {
+export function getFuntionSignatures(contract: Contract): string[] {
```
ItsNickBarry: Spelling

0xCourtney: this was fixed in https://github.com/solidstate-network/solidstate-solidity/pull/158/commits/dbf11d58067b3f5b081ef0d26414905f825fc4c1.

ItsNickBarry: Clicking the notification only showed this commit.  I feel like it used to show all new commits.

ItsNickBarry: Maybe I had the tab open since before you fixed?

0xCourtney: Probably because I edited the comment after fixing it

ItsNickBarry: ha, okay

jhubbardsf:
Hey, anything happening with this PR? Does anyone need some help and getting it finished? I saw a lot of promise in it.

ItsNickBarry:
@jhubbardsf  I've shifted focus to a WIP plugin that manages diamond upgrades.  It's easier to add the functions I need there first, and then determine how to structure what's been added here.  This will be merged eventually.

PR #164, opened 2022-11-09
Update `UintUtils`

ItsNickBarry:
Several API changes and optimizations.

Further changes needed:
* [x] `toString` tests
* [x] `toBinString` implementation
* [ ] `toBase64String` implementation (see [here](https://github.com/Brechtpd/base64/blob/main/base64.sol))
* [x] reconcile padding input values - `toDecString(uint256,uint256)` accepts a string length, while `toHexString(uint256,uint256)` accepts a bytes length
* [x] prefixes (`0x`) and possibly hex length enforcement (`0x1` => `0x01`) should be optional, and other prefixes should be available (see `0b` and `0o` [here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigInt)) (edit: hex length enforcement is optional when using `toString` with base 16, but not when using `toHexString`)


ItsNickBarry:
@NouDaimon Okay to review now.  Opinions about the missing features?

test/utils/UintUtils.ts
```diff
@@ -90,6 +112,44 @@ describe('UintUtils', function () {
       });
 
       describe('reverts if', () => {
+        describe('reverts if', () => {
```
NouDaimon, line 135: Should remove nested `describe`

NouDaimon (approved):

ItsNickBarry:
I think the best way to deal with the prefixes is to add custom functions for conversion to each prefixed base: `toHexString`, `toBinString`, etc.  This could also preserve the behavior of `Address#toString` found in other libraries, where a bytes length is provided rather than a string length - this PR currently standardizes all `toString` functions to use the latter.

ItsNickBarry:
Above comment is implemented except for reverting the behavior of the `length` parameter of `toHexString`.

ItsNickBarry:
Two questions remain:

Should `toHexString(uint256)` zero-pad the output so that character length is always even and bytes are not split?
```solidity
toHexString(1) => 0x01
// or
toHexString(1) => 0x1
```

Should `toHexString(uint256,uint256)` treat the `length` parameter as a bytes length or a character length?
```solidity
toHexString(1, 1) => 0x01
toHexString(1, 2) => 0x0001
// or
toHexString(1, 1) => 0x1
toHexString(1, 2) => 0x01
```

ItsNickBarry:
For both questions mentioned above, `length` shall refer to bytes length rather than character length.  Hex nibbles shall not be split, except when using the lower level `toString` function.

PR #170, opened 2022-11-30
Add `DoublyLinkedList`

### LINKED ISSUE #34: Linked List
SolidState needs libraries for working with singly and doubly linked lists.

ItsNickBarry:
closes #34


ItsNickBarry:
The `LinkedList` and `Queue` libraries are essentially the same and should be merged together.

ItsNickBarry:
`Queue` has been renamed to `DoublyLinkedList`, and `LinkedList` has been removed in 6ad3f5185cbce74154cf3f634e8a816c87b07a34.  Check `LinkedList` again after tests are written for stylistic and optimization improvements.

ItsNickBarry:
Still to do:
* [ ] make final decisions about when to revert and when to return `false`
* [ ] copy bytes32 tests for address and uint types

ItsNickBarry:
Tested combing the `_asc` and `_desc` mappings into one, differentiated with a `bool`, but this appeared to universally use more gas.

contracts/data/DoublyLinkedList.sol
```diff
@@ -0,0 +1,428 @@
...
+pragma solidity ^0.8.0;
+
+/**
+ * @title Doubly linked list implementation with enumeration functions
+ */
+library DoublyLinkedList {
+    struct DoublyLinkedListInternal {
+        mapping(bytes32 => bytes32) _nextValues;
+        mapping(bytes32 => bytes32) _prevValues;
+    }
```
0xCourtney, line 12: As per our earlier discussion; it may be worth considering adding a `head` and `tail` pointer to the struct which point to the first and last items of the DLL, respectively. 

For Example:

```md
head <-> A <-> B <-> C <-> D <-> tail
```

In this case `head == A`, `tail == D`. If `D` is removed from the DLL, `tail == C`  


```md
head <-> A <-> B <-> C <-> tail
```

contracts/data/DoublyLinkedList.sol
```diff
@@ -0,0 +1,428 @@
...
+    ) internal view returns (address) {
+        return
+            address(
+                uint160(
+                    uint256(
+                        _prev(self._inner, bytes32(uint256(uint160(value))))
+                    )
+                )
+            );
+    }
```
0xCourtney, line 69: We convert the address to/from bytes32 a few times throughout SS. What do you think about adding a `toAddress` from `fromAddress` to the `AddressUtils`, or `SafeCast` library?

contracts/data/DoublyLinkedList.sol
```diff
@@ -0,0 +1,428 @@
...
+        address value
+    ) internal view returns (address) {
+        return
+            address(
+                uint160(
+                    uint256(
+                        _prev(self._inner, bytes32(uint256(uint160(value))))
+                    )
+                )
+            );
```
0xCourtney, line 68: We convert the address to/from bytes32 a few times throughout SS. What do you think about adding a `toAddress` from `fromAddress` to the `AddressUtils`, or `SafeCast` library?

ItsNickBarry: I thought the same thing, but this might necessitate adding an `UnsafeCast` library because within the data structures we can be sure that we're not losing data by casting, and don't want to waste gas by checking.

I think we'll do this at some point, but not in this PR.

contracts/data/DoublyLinkedList.sol
ItsNickBarry: In order for us to be able to query first/last element, we would have to store head/tail in a separate variable.  This means an additional storage write on every `push` (which I believe will be the most commonly used operation).

contracts/data/DoublyLinkedList.sol
```diff
@@ -0,0 +1,428 @@
...
+    function _prev(
+        DoublyLinkedListInternal storage self,
+        bytes32 nextValue
+    ) private view returns (bytes32 prevValue) {
+        prevValue = self._prevValues[nextValue];
+        if (
+            nextValue != 0 &&
+            prevValue == 0 &&
+            _next(self, prevValue) != nextValue
+        ) revert DoublyLinkedList__InvalidReference();
```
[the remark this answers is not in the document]
ItsNickBarry: Renamed to `NonExistentEntry` for consistency with some similar errors elsewhere.

0xCourtney (approved):

PR #171, opened 2022-11-30
Remove storage functions

ItsNickBarry:


ItsNickBarry:
@NouDaimon ERC165 changes are implemented in #173 rather than here.

contracts/access/ownable/OwnableInternal.sol
```diff
@@ -26,22 +25,25 @@ abstract contract OwnableInternal is IOwnableInternal {
...
 
     function _transferOwnership(address account) internal virtual {
-        OwnableStorage.layout().setOwner(account);
-        emit OwnershipTransferred(msg.sender, account);
+        _setOwner(account);
+    }
+
+    function _setOwner(address account) internal virtual {
+        OwnableStorage.Layout storage l = OwnableStorage.layout();
+        emit OwnershipTransferred(l.owner, account);
```
NouDaimon, line 46: Should the event be emitted after the state change?

contracts/token/ERC721/base/ERC721BaseInternal.sol
```diff
@@ -64,13 +64,12 @@ abstract contract ERC721BaseInternal is IERC721BaseInternal {
 
     function _mint(address to, uint256 tokenId) internal virtual {
         if (to == address(0)) revert ERC721Base__MintToZeroAddress();
-
-        ERC721BaseStorage.Layout storage l = ERC721BaseStorage.layout();
-
-        if (l.exists(tokenId)) revert ERC721Base__TokenAlreadyMinted();
+        if (_exists(tokenId)) revert ERC721Base__TokenAlreadyMinted();
```
NouDaimon, line 67: Would such functions benefit from having a layout parameter?

This would remove the need of more than one call of `layout()` in a single function.

Not sure if gas is more important than clarity here.

NouDaimon (approved):

contracts/token/ERC721/base/ERC721BaseInternal.sol
ItsNickBarry: No, I'm avoiding passing `l` as a parameter as a rule.  We used in in the Premia contracts because at the time our architecture required contract size savings, but there's always a way around that when necessary.

contracts/access/ownable/OwnableInternal.sol
ItsNickBarry: If this were a large, complex function, it might be worth it to store `l.owner` in stack so that it can be emitted in the event at the end.  But it's clear enough what's happening here without adding the stack variable.

PR #176, opened 2022-12-20
Use `FacetCutAction` enum in tests

ItsNickBarry:
@0xCourtney Not sure why this isn't working.  Can you check?


ItsNickBarry:
Fixed with `yarn run lerna bootstrap`.

PR #178, opened 2022-12-25
Add virtual modifier to ERC721Base functions

NouDaimon:
Recently OpenSea has announced the implementation of an operator filter registry: https://github.com/ProjectOpenSea/operator-filter-registry 

This will prevent collections from receiving creator fees if OpenSea cannot confirm compliance with the above, or if it is found that they are trading assets on non-approved addresses, by making creator fees optional on OpenSea:

"Starting January 2nd, 2023, Opensea will begin validating creator fee enforcement for new collections on all supported EVM chains. After January 2nd, 2023, if OpenSea is unable to validate enforcement, OpenSea will make creator fees optional for that collection. Older collections will continue to have their fees enforced on OpenSea, including on Ethereum Mainnet (previously, enforcement was already required on Ethereum Mainnet)."

Compliance with OpenSea requires overriding of basic ERC721 functions. It is likely that if SolidState-based ERC721 collections wish to trade on OpenSea, they will require the ability to override the `ERC721Base` to abide by the `Operator-filter-registry`.

### DIFF
--- a/contracts/token/ERC20/base/ERC20Base.sol
+++ b/contracts/token/ERC20/base/ERC20Base.sol
@@ -14,14 +14,14 @@ abstract contract ERC20Base is IERC20Base, ERC20BaseInternal {
     /**
      * @inheritdoc IERC20
      */
-    function totalSupply() public view returns (uint256) {
+    function totalSupply() external view returns (uint256) {
         return _totalSupply();
     }
 
     /**
      * @inheritdoc IERC20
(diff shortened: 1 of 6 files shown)


ItsNickBarry:
@NouDaimon Fixed in a different way as discussed elsewhere.  Please review and merge.

ItsNickBarry (approved):

PR #185, opened 2023-01-13
event util

### LINKED ISSUE #61: Event parsing function
Draft provided by @Lumyo 

```typescript
import { ContractReceipt } from 'ethers';
import { Interface } from 'ethers/lib/utils';
import { LogDescription } from '@ethersproject/abi';

function getEvents(
  contractInterface: Interface,
  receipt: ContractReceipt,
) {
  const events: LogDescription[] = [];

  for (const log of receipt.logs) {
    try {
      events.push(contractInterface.parseLog(log));
    } catch (e) {}
  }

  return events;
}
```

0xCourtney:
closes #61

### DIFF
--- a/lib/event.ts
+++ b/lib/event.ts
@@ -17,7 +17,7 @@ export function getLogs(
         continue;
       }
 
-      events.push(contractInterface.parseLog(log));
+      events.push(description);
     } catch (e) {
       console.log(e);
       throw Error('could not parse logs');
--- a/lib/event.ts
+++ b/lib/event.ts
@@ -11,13 +11,13 @@ export function getLogs(
 
   for (const log of receipt.logs) {
     try {
-      const description = contractInterface.parseLog(log);
+      const parsedLog = contractInterface.parseLog(log);
 
-      if (only.length > 0 && !only.includes(description.name)) {
+      if (only.length > 0 && !only.includes(parsedLog.name)) {
         continue;
       }
 
-      events.push(description);
+      events.push(parsedLog);
     } catch (e) {
       console.log(e);
       throw Error('could not parse logs');
--- a/lib/event.ts
+++ b/lib/event.ts
@@ -1,0 +1,28 @@
+import { LogDescription } from '@ethersproject/abi';
+import { ContractReceipt } from 'ethers';
+import { Interface } from 'ethers/lib/utils';
+
+export function getLogs(
+  contractInterface: Interface,
+  receipt: ContractReceipt,
+  only: string[] = [],
+) {
+  const events: LogDescription[] = [];
+
+  for (const log of receipt.logs) {
+    try {
+      let description = contractInterface.parseLog(log);
+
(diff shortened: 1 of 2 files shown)


lib/event.ts
```diff
@@ -0,0 +1,28 @@
+import { LogDescription } from '@ethersproject/abi';
+import { ContractReceipt } from 'ethers';
+import { Interface } from 'ethers/lib/utils';
+
+export function getLogs(
+  contractInterface: Interface,
+  receipt: ContractReceipt,
+  only: string[] = [],
```
ItsNickBarry, line 8: Might as well add `except` here to continue the only/except pattern.  But I'm thinking about making a separate package to handle that logic because it's used it so many places, so we can just wait for that.

lib/event.ts
```diff
@@ -0,0 +1,28 @@
...
+) {
+  const events: LogDescription[] = [];
+
+  for (const log of receipt.logs) {
+    try {
+      let description = contractInterface.parseLog(log);
```
ItsNickBarry, line 14:
```suggestion
      const description = contractInterface.parseLog(log);
```

lib/event.ts
```diff
@@ -0,0 +1,28 @@
...
+
+      if (only.length > 0 && !only.includes(description.name)) {
+        continue;
+      }
+
+      events.push(contractInterface.parseLog(log));
```
ItsNickBarry, line 20: Already parsed, so just push `description`.
```suggestion
      events.push(description);
```

ItsNickBarry (changes requested):

ItsNickBarry:
@0xCourtney Is this something you want published now, or are you just trying to resolve the open issue?

0xCourtney:
> @0xCourtney Is this something you want published now, or are you just trying to resolve the open issue?

This is not urgent. I was thinking about this a bit recently and thought it would be good to share what I worked on.

lib/event.ts
0xCourtney: Absolutely, the filter I did for Diamond utils also supports `except` so it may make the most sense to wait until that's merged or branch from that PR and break out the filters specifically.

lib/event.ts
```diff
@@ -0,0 +1,28 @@
...
+export function getLogs(
+  contractInterface: Interface,
+  receipt: ContractReceipt,
+  only: string[] = [],
+) {
+  const events: LogDescription[] = [];
+
+  for (const log of receipt.logs) {
+    try {
+      const description = contractInterface.parseLog(log);
```
ItsNickBarry, line 14: I don't like this variable name.  Can change it in a future reorganization, though.

ItsNickBarry (approved):

lib/event.ts
0xCourtney: renamed this to `parsedLog`

ItsNickBarry (approved):
Merge if ready.

PR #188, opened 2023-01-26
Add `EIP712` library

ItsNickBarry:


0xCourtney (approved):

test/cryptography/EIP712.ts
```diff
@@ -0,0 +1,49 @@
...
+        const chainId = await ethers.provider.send('eth_chainId');
+
+        const domainSeparator = ethers.utils.solidityKeccak256(
+          ['bytes32', 'bytes32', 'bytes32', 'uint256', 'address'],
+          [
+            typeHash,
+            nameHash,
+            versionHash,
+            chainId,
+            ethers.utils.hexZeroPad(instance.address, 32),
```
ItsNickBarry, line 36: @0xCourtney I don't understand why the `hexZeroPad` is needed here.  Any idea?

Lumyo: `solidityKeccak256` is packing values. 

This would work :
```ts
        import { defaultAbiCoder, keccak256 } from 'ethers/lib/utils';

        const domainSeparator = keccak256(defaultAbiCoder.encode(['bytes32', 'bytes32', 'bytes32', 'uint256', 'address'], [
            typeHash,
            nameHash,
            versionHash,
            chainId,
            instance.address,
          ]));
```

0xCourtney: In solidity, the domain is constructed without packed encoding (ie `abi.encode`), according to the `ethers` [docs](https://docs.ethers.org/v5/api/utils/hashing/#utils-solidityKeccak256) `solidityKeccak256` packs encoded values, as @Lumyo mentioned.

> Returns the [KECCAK256](https://en.wikipedia.org/wiki/SHA-3) of the non-standard encoded values packed according to their respective type in types.

We should probably use `keccak256` in this test to be consistent. 

Screenshot 2023-05-24 at 11 16 24

ItsNickBarry:
Published in `v0.0.57`.

PR #189, opened 2023-01-28
Inherit `OwnableBehaviorArgs` into `DiamondWritableBehaviorArgs`

ItsNickBarry:
Sharing these behavior arguments is probably ideal in most situations, but it might be more strictly correct to define new arguments `diamondCutter` and `nonDiamondCutter`, in case that role is not the same as the `owner` role.


ItsNickBarry:
Might address the concerns above in a future PR, but it's not important.

PR #192, opened 2023-02-14
Comment for Token Implementers

zdenham:
Small comment that would have saved a bit of time sleuthing when implementing the token standards:
BaseImplementation


ItsNickBarry (approved):
Though it's not clear from the inline IDE warning, the compilation error should point to the exact issue.  However, additional documentation is usually not bad, so I'll merge.

jhubbardsf:
Yeah, the compilation message is pretty clear.
```
TypeError: Contract "ERC1155Facet" should be marked as abstract.
  --> contracts/ERC1155-Diamond/facets/ERC1155Facet.sol:26:1:
   |
26 | contract ERC1155Facet is ERC1155Base {
   | ^ (Relevant source part starts here and spans across multiple lines).
Note: Missing implementation:
  --> @solidstate/contracts/interfaces/IERC165.sol:15:5:
   |
15 |     function supportsInterface(bytes4 interfaceId) external view returns (bool);
```

PR #193, opened 2023-02-17
`FacetCut` struct  in `IDiamondWriteableInternal` doesn't follow the Spec naming conventions

zdenham:
The naming convention for the `FacetCut` struct does not follow the diamond [spec](https://eips.ethereum.org/EIPS/eip-2535), which uses a different naming convention `facetAddress` and `functionSelectors`:

https://github.com/solidstate-network/solidstate-solidity/blob/ae15a063728978ed30ddfd2b3faedb775307c417/contracts/proxy/diamond/writable/IDiamondWritableInternal.sol#L23-L28

This could potentially break client compatibility across diamonds. E.G. when I was porting over an ethers `diamondCut` script from another project, there were type errors with the SS Diamond.

This PR updates the codebase and tests to match the standard naming conventions.

I actually like the SS naming convention better, but it might be worth maintaining consistency with the standard.

**EDIT**: this also isn't a problem for everyone: I tested with louper, and their `diamondCut` still works with SS diamond despite using the spec naming convention, *because the ABI for the diamond they are referencing uses the same convention.

https://github.com/mark3labs/louper-v2/blob/d326c9ba4c48724297a6b8f5dcc9915d148ed250/src/lib/components/AddFacet.svelte#L41-L47

### DIFF
--- a/abi/DiamondWritable.json
+++ b/abi/DiamondWritable.json
@@ -56,7 +56,7 @@
         "components": [
           {
             "internalType": "address",
-            "name": "target",
+            "name": "facetAddress",
             "type": "address"
           },
           {
@@ -66,7 +66,7 @@
           },
           {
             "internalType": "bytes4[]",
-            "name": "selectors",
+            "name": "functionSelectors",
             "type": "bytes4[]"
           }
         ],
@@ -116,7 +116,7 @@
         "components": [
           {
             "internalType": "address",
-            "name": "target",
+            "name": "facetAddress",
             "type": "address"
           },
           {
@@ -126,7 +126,7 @@
           },
           {
             "internalType": "bytes4[]",
-            "name": "selectors",
+            "name": "functionSelectors",
             "type": "bytes4[]"
           }
         ],
--- a/abi/DiamondWritableInternal.json
+++ b/abi/DiamondWritableInternal.json
@@ -46,7 +46,7 @@
         "components": [
           {
             "internalType": "address",
-            "name": "target",
+            "name": "facetAddress",
             "type": "address"
           },
           {
@@ -56,7 +56,7 @@
           },
           {
             "internalType": "bytes4[]",
-            "name": "selectors",
+            "name": "functionSelectors",
             "type": "bytes4[]"
           }
         ],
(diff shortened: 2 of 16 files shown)


ItsNickBarry:
Sorry for the delay.

My first response was to say that the variable names are not part of the spec because they have no effect on ABIs.  I disagree with several naming choices in EIP 2535 (as well as most other EIPs) and have made changes where possible.  However, I've been simultaneously looking at EIP-712 (#188), which disturbingly uses variable names in its signature generation scheme.

So, I'm torn between making use of the freedom which should be allowed by the ABI system, and trying to accommodate an EIP that pretends like such freedom does not exist.

zdenham:
Forgot about this 😅 

> variable names [...] have no effect on ABIs. 

The encoded payload ends up being the same, but when updating a struct variable name and re-compiling the build artifacts / ABI are modified accordingly, which affects generated typechain types and validation at ethers.js level.

This was a minor inconvenience for me because I had a strongly typed `diamondCut` function (in typescript) as a part of a small internal library that broke when using SS ABIs because it expected "standard" naming convention. Was a design flaw of the lib mostly though, and a quick fix. But that is where the idea to update this came from.

That being said, if no one else has mentioned / run into similar issues since its probably not worth changing. Down to close this for cleanliness

ItsNickBarry:
I think I'm going to end up merging.  I'm just having a hard time accepting the implications.

ItsNickBarry:
> I think I'm going to end up merging. I'm just having a hard time accepting the implications.

I've used EIP-712 a few times now, and have come to the conclusion that contracts that implement it will always have to explicitly define their type hashes.  If the EIP were designed differently, with hashes generated from unnamed parameter types, this process could have been automatic.  There's no point in trying to standardize parameter names for the sake of the EIP.

(Not relevant, but it occurs to me that a further EIP could fix EIP-712 by standardizing the parameter names as `p1`, `p2`, `p3`, etc.  This would make automatic type hash generation possible.)

Therefore, I've decided not to merge this.  Parameter names are not standard, so tools must not treat them as such.  Obviously you've forgotten about this, but thanks for the PR, @zdenham.

zdenham:
haha you are correct, I'd forgotten about this one. But thanks for the follow up anyhow.

Glad to see these contracts are still well maintained. Keep going 🫡

PR #197, opened 2023-03-05
Internal `ReentrancyGuard` functions

ItsNickBarry:
Slight increase in gas cost unless compiling with `viaIR` option.

New internal functions need tests.


ItsNickBarry:
The code from this PR has been included in #199 and modified in such a way that it does not work as originally intended.  Published in `v0.0.55`.

PR #209, opened 2023-04-01
No cyclic dependencies

### LINKED ISSUE #208: Help with Flattening Files
I have a simple ERC721 Contract and 
```
// SPDX-License-Identifier: MIT
pragma solidity ^0.8.9;
import "@solidstate/contracts/token/ERC721/SolidStateERC721.sol";

contract SampleERC721 is SolidStateERC721 {
  //Contract data.....
}
```
The contract compiles properly except when I try to Flatten it I get this error. 

Error: **There is a cycle in the dependency graph, can't compute topological ordering.**

I don't understand the error. I know there is a loop that causes it to trigger as a parent is also acting like a child or something like that. 

Can you tell me how to I flatten it?

ItsNickBarry:
closes #208

### DIFF
--- a/contracts/data/IncrementalMerkleTreeMock.sol
+++ b/contracts/data/IncrementalMerkleTreeMock.sol
@@ -4,8 +4,6 @@ pragma solidity ^0.8.0;
 
 import { IncrementalMerkleTree } from './IncrementalMerkleTree.sol';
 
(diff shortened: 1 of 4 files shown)


ItsNickBarry:
Merging without review because only tests and imports are affected.

PR #216, opened 2023-05-23
Add enumerable functions to AccessControl

tess3rac7:
We add two new function, getRoleMember(role, index) and getRoleMemberCount(role) to AccessControl. The function and parameter names are borrowed from OpenZeppelin to make them familiar. These are quite lightweight, and AccessControlStorage already uses EnumerableSet, so it made sense to just add them to AccessControl instead of creating a new AccessControlEnumerable contract. We also add tests for positive and revert cases.

### DIFF
--- a/contracts/access/access_control/IAccessControl.sol
+++ b/contracts/access/access_control/IAccessControl.sol
@@ -45,4 +45,27 @@ interface IAccessControl is IAccessControlInternal {
      * @param role role to relinquish
      */
     function renounceRole(bytes32 role) external;
+
+    /**
+     * @notice Returns one of the accounts that have `role`. `index` must be a
+     * value between 0 and {getRoleMemberCount}, non-inclusive.
+     *
+     * Role bearers are not sorted in any particular way, and their ordering may
+     * change at any point.
+     *
+     * WARNING: When using {getRoleMember} and {getRoleMemberCount}, make sure
+     * you perform all queries on the same block. See the following
+     * https://forum.openzeppelin.com/t/iterating-over-elements-on-enumerableset-in-openzeppelin-contracts/2296[forum post]
+     * for more information.
+     */
+    function getRoleMember(
+        bytes32 role,
+        uint256 index
+    ) external view returns (address);
+
+    /**
+     * @notice Returns the number of accounts that have `role`. Can be used
+     * together with {getRoleMember} to enumerate all bearers of a role.
+     */
+    function getRoleMemberCount(bytes32 role) external view returns (uint256);
 }
--- a/test/access/AccessControl.ts
+++ b/test/access/AccessControl.ts
@@ -14,10 +14,12 @@ const ROLE = ethers.utils.solidityKeccak256(['string'], ['ROLE']);
 describe('AccessControl', function () {
   let admin: SignerWithAddress;
   let nonAdmin: SignerWithAddress;
+  let nonAdmin2: SignerWithAddress;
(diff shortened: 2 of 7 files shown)


lib/package-lock.json
tess3rac7, line 1: I only used `yarn`, not sure why this was generated. Happy to remove from the PR if needed, please let me know.

spec/package-lock.json
tess3rac7, line 1: I only used `yarn`, not sure why this was generated. Happy to remove from the PR if needed, please let me know.

yarn.lock
```diff
@@ -1983,7 +1983,7 @@
   dependencies:
     axios "^0.24.0"
 
-"@solidstate/library@^0.0.54":
+"@solidstate/library@^0.0.56":
```
tess3rac7, line 1986: I did not run any specific commands to update this. Happy to remove from the PR if needed. Please let me know.

test/access/AccessControl.ts
```diff
@@ -103,4 +105,57 @@ describe('AccessControl', function () {
...
+      await instance.connect(admin).grantRole(ROLE, nonAdmin2.address);
+      await instance.connect(admin).grantRole(ROLE, nonAdmin3.address);
+
+      expect(await instance.getRoleMember(ROLE, 0)).to.equal(nonAdmin.address);
+      expect(await instance.getRoleMember(ROLE, 1)).to.equal(nonAdmin2.address);
+      expect(await instance.getRoleMember(ROLE, 2)).to.equal(nonAdmin3.address);
+    });
+
+    describe('reverts if', function () {
+      it('role does not exist', async function () {
```
tess3rac7, line 121: I suppose `EnumerableSet` doesn't care about the role existing. But it felt idiomatic from the `AccessControl` POV to test both cases (role doesn't exist + role exist but index is invalid). But happy to collapse into a single test if needed, please let me know.

contracts/access/access_control/IAccessControl.sol
```diff
@@ -45,4 +45,27 @@ interface IAccessControl is IAccessControlInternal {
...
+     * https://forum.openzeppelin.com/t/iterating-over-elements-on-enumerableset-in-openzeppelin-contracts/2296[forum post]
+     * for more information.
+     */
+    function getRoleMember(
+        bytes32 role,
+        uint256 index
+    ) external view returns (address);
+
+    /**
+     * @dev Returns the number of accounts that have `role`. Can be used
```
ItsNickBarry, line 67: Use `notice` tags for high-level descriptions and `dev` tags for obscure developer notes.  I don't know if this is the "correct" usage of these tags, but it's how it's done in the rest of SS.

test/access/AccessControl.ts
ItsNickBarry: Seems fine as-is.

ItsNickBarry (changes requested):
> AccessControlStorage already uses EnumerableSet, so it made sense to just add them to AccessControl instead of creating a new AccessControlEnumerable contract

This is true, but see ERC721: the base contract is made a bit less inefficient in order to accommodate the separate enumerable extension.  In that case it was done so that the base contract corresponds 1:1 with the ERC721 standard interface.

I agree that access control doesn't need a separate extension, but might change this in the future.

tess3rac7:
Updated :slightly_smiling_face:

ItsNickBarry (approved):
Thanks for the contribution.  Will merge and publish soon.

Issue #85, opened 2022-03-30
Suggest removing `diamondCut` as an immutable function


mudgen:
From the EIP2535 Diamonds standard an immutable function is a function defined directly in a diamond so it can never be removed.

In solidstate-solidity the `diamondCut` function is defined as an immutable function in the Diamond.sol implementation: https://github.com/solidstate-network/solidstate-solidity/blob/master/contracts/proxy/diamond/Diamond.sol

But one of the features of a diamond is the ability to remove the `diamondCut` function at some point to make an immutable diamond.  It would be useful if the `diamondCut` function was added to a diamond, but not as an immutable function so diamond owners have the ability to remove it or replace it.

ItsNickBarry:
`Diamond.sol` is the "SolidState recommended implementation" of a diamond proxy.  Most users will likely never want to remove the `diamondCut` function, so it is marked as immutable here.  Users who do plan to "finalize" a diamond at some point can use `DiamondBase.sol` to set up their own diamonds, and add a contract derived from `DiamondCuttable.sol` as a non-immutable facet.

I intend to describe the difference between base components and "SolidState recommended" contracts in a future documentation update.  As an example, `ERC20.sol` works the same way - it is an "opinionated" contract in that it includes additional approval and metadata functions which are not part of the ERC20 standard but are desirable in most cases.

PR #219, opened 2023-06-13
Make `_diamondCut` virtual

zdenham:
Motivation is to give users a path to implement  "finalizing" their diamond  (make them immutable) without having to eject from `SolidStateDiamond`, which is a very useful and easy-to-use default.

Related to https://github.com/solidstate-network/solidstate-solidity/issues/85.  I personally feel the desire to finalize diamonds may be more common than a power user feature. 

Making `diamondCut` virtual allows for the ability to conditionally revert on diamondCut if a "UpgradeabilityRevoked" flag is set somewhere, or other user specific conditions to finalize a diamond e.g. mutability expiration date etc...

### DIFF
--- a/contracts/proxy/diamond/writable/DiamondWritableInternal.sol
+++ b/contracts/proxy/diamond/writable/DiamondWritableInternal.sol
@@ -24,7 +24,7 @@ abstract contract DiamondWritableInternal is IDiamondWritableInternal {
         FacetCut[] memory facetCuts,
         address target,
         bytes memory data
-    ) internal {
+    ) internal virtual {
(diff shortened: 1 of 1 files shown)


zdenham:
Example implementation could be 

```solidity
contract MyDiamondWithRevokableUpgradeability is SolidStateDiamond {
...
    function _diamondCut(
        FacetCut[] memory facetCuts,
        address target,
        bytes memory data
    ) internal override {
      require(!isUpgradeabilityRevoked, "upgradeability revoked");
      super._diamondCut(...args);
    }
}
```

ItsNickBarry (approved):
This should have been `virtual` anyway, because it's an `internal` function.

PR #220, opened 2023-06-13
Upgrade to Ethers V6

ItsNickBarry:
This branch upgrades ethers from V5 to V6.  Waffle has not been updated, so an alternative to Waffle mock contracts might be needed before merging.

also closes #148

PR #225, opened 2023-07-28
Fix Prettier

ItsNickBarry:
Prettier plugin handling seems to have changed during development of #220.  This branch properly imports both plugins in use.  However, CLI commands seem to ignore the `plugins` field of `.prettierrc.json`, so the plugins must be redundantly included in the yarn script.

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,1 +1,0 @@
-node_modules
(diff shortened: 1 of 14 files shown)


ItsNickBarry:
Looks like the plugin issue is being fixed here: https://github.com/prettier/prettier/pull/15155

ItsNickBarry:
Merging without review because only configuration and test files are affected.

PR #226, opened 2023-07-28
Move delegatecall logic from fallback function

0xCourtney:


ItsNickBarry:
The `_fallback` function should not use `AddressUtils` because of the extra gas overhead.  I've reverted that to inline delegatecall, but we can keep the new `AddressUtils` functions.

PR #232, opened 2023-11-21
Upgrade minimum version of solc to ^0.8.18.

Hotmanics:
[Slither](https://github.com/crytic/slither) is becoming an increasingly useful tool for analyzing smart contracts and recognizing vulnerabilities within them. 

When running this tool against the smart contracts in this codebase, it returns several vulnerabilities pertaining to the solidity compiler version used.

Screenshot 2023-11-21 at 1 42 41 AM

Namely, v0.8.8 contains this bug ([documented here](https://docs.soliditylang.org/en/latest/bugs.html)):

```
{
   "uid": "SOL-2021-4",
    "name": "UserDefinedValueTypesBug",
    "summary": "User defined value types with underlying type shorter than 32 bytes used incorrect storage layout and wasted storage",
    "description": "The compiler did not correctly compute the storage layout of user defined value types based on types that are shorter than 32 bytes. It would always use a full storage slot for these types, even if the underlying type was shorter. This was wasteful and might have problems with tooling or contract upgrades.",
    "link": "https://blog.soliditylang.org/2021/09/29/user-defined-value-types-bug/",
    "introduced": "0.8.8",
    "fixed": "0.8.9",
    "severity": "very low"
},
```

Hence, we can get rid of the warnings from slither and enhance the security of projects that implement this codebase by upgrading the smart contracts to require a more secure version of the solidity compiler (I.E. the proposed 0.8.18 found within these commits).

### DIFF
--- a/contracts/access/access_control/AccessControl.sol
+++ b/contracts/access/access_control/AccessControl.sol
@@ -1,6 +1,6 @@
 // SPDX-License-Identifier: MIT
 
-pragma solidity ^0.8.0;
+pragma solidity ^0.8.9;
 
 import { IAccessControl } from './IAccessControl.sol';
 import { AccessControlInternal } from './AccessControlInternal.sol';
--- a/contracts/access/access_control/AccessControl.sol
+++ b/contracts/access/access_control/AccessControl.sol
@@ -1,6 +1,6 @@
 // SPDX-License-Identifier: MIT
 
-pragma solidity ^0.8.9;
+pragma solidity ^0.8.18;
 
 import { IAccessControl } from './IAccessControl.sol';
 import { AccessControlInternal } from './AccessControlInternal.sol';
--- a/contracts/access/access_control/AccessControlInternal.sol
+++ b/contracts/access/access_control/AccessControlInternal.sol
@@ -1,6 +1,6 @@
 // SPDX-License-Identifier: MIT
 
-pragma solidity ^0.8.9;
+pragma solidity ^0.8.18;
 
 import { EnumerableSet } from '../../data/EnumerableSet.sol';
 import { AddressUtils } from '../../utils/AddressUtils.sol';
--- a/contracts/access/access_control/AccessControlInternal.sol
+++ b/contracts/access/access_control/AccessControlInternal.sol
@@ -1,6 +1,6 @@
 // SPDX-License-Identifier: MIT
 
-pragma solidity ^0.8.0;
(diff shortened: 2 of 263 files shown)


ItsNickBarry:
Old versions are allowed by design.  Imagine a situation where a developer wants to use SS alongside a third-party library which is tied to a single version.  It's preferable to allow the developer to choose a version that works with both libraries, with the understanding that they take responsibility for verifying that the chosen version is safe.  If SS is strictly incompatible with some third-party libraries, the developer is probably going to copy/paste code and update the pragma rather than import, which is generally worse.

With that said, I don't think we need to support versions as old as `0.8.0`.  Version `0.8.8` is set for files which use features introduced in that version, but this is also an arbitrary choice.  Is there a particular reason you chose `0.8.18`?

ItsNickBarry:
Nobody has indicated that this breaking change would inconvenience them, so I'm willing to merge it.  But first I'd like to add a test to ensure that the pragma statement remains consistent across all files, especially when outstanding pull requests are merged.  It would look sort of like the tests in `internal.ts` and `sources.ts`.

Hotmanics:
> Nobody has indicated that this breaking change would inconvenience them, so I'm willing to merge it. But first I'd like to add a test to ensure that the pragma statement remains consistent across all files, especially when outstanding pull requests are merged. It would look sort of like the tests in `internal.ts` and `sources.ts`.

Apologies, I had meant to reply to your original message.

The choice for 0.8.18 was for the sake of satisfying the slither check. It sees that version as the most stable/secure. I imagine as time goes on that this will change and we will need to update this repository accordingly. This shouldn't usually shouldn't require breaking changes, but will be a pain to keep up to date all the time. 

However, I do see your explanation for allowing older versions to be useful. When I worked in Unity, I made sure to support as many versions as possible (Just like solidity, the game engine itself had like 30+ version). I believe with Solidity it is a bit different though as the security obtained from restricting to newer versions would be a large enough benefit. Still new to OSS, so I don't know the best plan of action. Maybe there is a way to appease both sides of aisle?

ItsNickBarry:
I added the test mentioned above.  Need to find a place to define the `0.8.18` value other than in the test itself.  It still feels very arbitrary.

ItsNickBarry:
Modified the test to avoid the issue mentioned above.  I'll merge this, but no guarantee that we'll stay up-to-date with the Slither-recommended Solidity version in the future.  Thanks for the contribution.

PR #233, opened 2023-12-01
Beacon patterns

ItsNickBarry:
This branch replaces the `ManagedProxy` contracts with a standard "beacon" pattern, and introduces a `DiamondBeacon`.  The latter is an improvement over the pattern used in the Premia and Insrt contracts that has been frequently recommended as a means of deploying diamond-like contracts via factory.


ItsNickBarry:
Is the `Beacon`/`BeaconProxy` naming pattern easy to understand?  Would it be better to use something like `BeaconEmitter`/`BeaconReceiver`?

ItsNickBarry:
Diamond initialization feature must be removed from `DiamondBeacon#diamondCut`.

PR #236, opened 2024-01-30
Implement ERC7201

### LINKED ISSUE #162: Allow custom storage slots
A function like this must be added to each storage library:

```solidity
function layout(bytes32 slot) internal pure returns (Layout storage l) {
    assembly {
        l.slot := slot
    }
}
```

### LINKED ISSUE #166: Use EIP-7201 to define storage slots
[This EIP](https://eips.ethereum.org/EIPS/eip-1967) uses `keccak256` to generate storage slots, and subtracts 1 from the result to avoid preimage attacks.  This works for the EIP because it uses only single-slot storage values, but not for SolidState because many of the storage layout structs are multi-slot.

Should try to address this?  It's not entirely clear what such a preimage attack would look like.  If such an attack can be demonstrated, I would suggest using bitwise-not rather than subtraction of 1 to obscure the SolidState storage slots.

ItsNickBarry:
closes #166
closes #162


ItsNickBarry:
Include a test that checks each storage library for correct storage slot declaration, so that other conflicting branches can be easily merged.

ItsNickBarry:
Should add the custom annotation to the storage library `Layout` structs:
```
    /**
     * @custom:storage-location erc7201:solidstate.contracts.storage.ContractName
     */
```

ItsNickBarry:
Before merging, need to decide whether to keep the storage libraries within the scope of the layers pattern (same base name as other layers, base name used as part of storage slot calculation), or associate them more closely with their EIPs or higher level intentions.

For example:
* In the former case, we keep awkward names like `ERC20Base` and use that string to define the storage slot permanently.
* In the latter case, related storage libraries could be combined: `ERC20Base`, `ERC20Metadata` => `ERC20`

PR #237, opened 2024-01-31
update minimum pragma version to 0.8.20

ItsNickBarry:
Merging this change directly into `master` to make the diff of #236 easier to read.


ItsNickBarry:
I don't remember why the linked PR requires this solc version, but it's been out long enough that I'm going to merge now, ending support for solc < `0.8.20`.

PR #238, opened 2024-02-03
Diamond `receive` routing

ItsNickBarry:
Basic test of support for dynamic `receive` function in diamond proxies.


ItsNickBarry:
Merging directly because these are just tests.

PR #239, opened 2024-02-25
replace `safeApprove` with `forceApprove`

ItsNickBarry:
The existing `safeApprove` function is redundant at best (for reasons explained in the `@dev` natspec tag) and confusing at worst.  This branch removes the zero-balance check and renames the function to `forceApprove` to prevent silent failure for users who update Solidstate.  The name may be reverted to `safeApprove` in a future update, for consistency with the other `SafeERC20` functions.

The `SafeERC20` library does not have any tests, so thorough review is required.


ItsNickBarry:
@startswithaj

PR #247, opened 2024-05-22
Internal Diamond proxy cleanup

ItsNickBarry:
Refactoring + comments.  Branches merged into this branch make some behavior changes.


ItsNickBarry:
I don't know whether "slug" is the right terminology.

contracts/proxy/diamond/base/DiamondBaseStorage.sol
```diff
@@ -7,12 +7,12 @@ pragma solidity ^0.8.20;
...
     struct Layout {
-        // function selector => (facet address, selector slot position)
-        mapping(bytes4 => bytes32) facets;
+        // function selector => (facet address, selector slug position)
+        mapping(bytes4 => bytes32) selectorInfo;
         // total number of selectors registered
         uint16 selectorCount;
-        // array of selector slots with 8 selectors per slot
-        mapping(uint256 => bytes32) selectorSlots;
+        // array of 32-byte slugs with 8 selectors each
```
startswithaj, line 14: Could this be?

```
// array of Bytes32 slugs. Each Bytes32 contains 8, 4byte selectors.
```

contracts/proxy/diamond/writable/DiamondWritableInternal.sol
```diff
@@ -95,123 +103,124 @@ abstract contract DiamondWritableInternal is IDiamondWritableInternal {
...
-                            ~(CLEAR_SELECTOR_MASK >>
-                                oldSelectorInSlotPosition)) |
-                        (bytes32(lastSelector) >> oldSelectorInSlotPosition);
+                // overwrite the selector being deleted with the last selector in the array
 
-                    // update storage with the modified slot
-                    l.selectorSlots[oldSelectorsSlotCount] = oldSelectorSlot;
+                if (slugIndex == lastSlugIndex) {
+                    // selector is being removed from the last slug, which has already been loaded
+                    // slug needs not be written to storage because it continues to be tracked
```
startswithaj, line 198: I think this comment should be made clearer

contracts/proxy/diamond/writable/DiamondWritableInternal.sol
```diff
@@ -263,4 +272,14 @@ abstract contract DiamondWritableInternal is IDiamondWritableInternal {
             }
         }
     }
+
+    function _overwriteSelector(
```
startswithaj, line 276: `_overwriteSelectorInSlug`?

contracts/proxy/diamond/readable/DiamondReadableInternal.sol
```diff
@@ -102,23 +102,23 @@ abstract contract DiamondReadableInternal is IDiamondReadableInternal {
...
+                slugSelectorIndex++
             ) {
                 selectorIndex++;
 
                 if (selectorIndex > l.selectorCount) {
                     break;
                 }
 
-                bytes4 selector = bytes4(slot << (selectorSlotIndex << 5));
+                bytes4 selector = bytes4(slug << (slugSelectorIndex << 5));
```
startswithaj, line 119: Maybe while we are adding comments, we should add one here explaining whats happening?

startswithaj:
Looks good. Fully support the name changes and additional comments. Nice one.

contracts/proxy/diamond/writable/DiamondWritableInternal.sol
ItsNickBarry: Would this be resolved by the function name change you suggested below?

contracts/proxy/diamond/readable/DiamondReadableInternal.sol
ItsNickBarry: It's not necessarily out of scope, but because primary purpose of this PR is to avoid write errors, I haven't changed the read functions much except for where storage variable names have changed.  I'll add comments to this file if I have time before #248 is merged.  If not, it can be done in a future PR.

contracts/proxy/diamond/base/DiamondBaseStorage.sol
startswithaj: I think we use `array` because the slug is indexed by `uint256`. Were using the mapping like an array.

contracts/proxy/diamond/writable/DiamondWritableInternal.sol
startswithaj:
```

// the selector being removed is from the last slug, this means we replace it in place
// the slug will be written to storage upstream, when returned. 
```

ItsNickBarry: Made some changes here.  There's also another comment above the if/else blocks that applies to both cases.

contracts/proxy/diamond/base/DiamondBaseStorage.sol
ItsNickBarry: Yes.  I believe it's a mapping because an array's `length` would be updated in storage on each push or pop.  Might be more efficient to use an array anyway, but use assembly to avoid the length update.  Can look into this later.

PR #251, opened 2025-01-07
`Math#sqrt` fix for max `uint256` input

### LINKED ISSUE #250: `Math.sqrt` does not support input value 2 ^ 256 - 1
When [the aforementioned function](https://github.com/solidstate-network/solidstate-solidity/blob/master/contracts/utils/Math.sol#L48-L61) is called with input `x = type(uint256).max`, it reverts on overflow in the first line:
```
    function sqrt(uint256 x) internal pure returns (uint256 y) {
        uint256 z = (x + 1) >> 1;
        y = x;
        while (z < y) {
            y = z;
            z = (x / z + z) >> 1;
        }
    }
```
You can resolve it using the following implementation instead:
```
    function sqrt(uint256 n) internal pure returns (uint256) {
        if (n > 0) {
            uint256 x = n / 2 + 1;
            uint256 y = (x + n / x) / 2;
            while (x > y) {
                x = y;
                y = (x + n / x) / 2;
            }
            return x;
        }
        return 0;
    }
```
Note that you can also scope the entire contents of this function under `unchecked` in order to reduce its gas consumption (otherwise `>> 1` would be preferable over `/ 2`).

See [here](https://github.com/barakman/solidity-math-utils/blob/master/project/contracts/IntegralMath.sol#L33-L47) for reference.

ItsNickBarry:
closes #250

### DIFF
--- a/contracts/utils/Math.sol
+++ b/contracts/utils/Math.sol
@@ -48,15 +48,18 @@ library Math {
     /**
      * @notice estimate square root of number
      * @dev uses Babylonian method (https://en.wikipedia.org/wiki/Methods_of_computing_square_roots#Babylonian_method)
-     * @param x input number
-     * @return y square root
+     * @param n input number
      */
-    function sqrt(uint256 x) internal pure returns (uint256 y) {
-        uint256 z = (x + 1) >> 1;
-        y = x;
-        while (z < y) {
-            y = z;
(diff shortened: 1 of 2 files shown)


ItsNickBarry:
Thanks @barakman for finding this issue.

PR #252, opened 2025-02-20
Enforce layers pattern

ItsNickBarry:
This PR adds tests to help avoid the errors in #196.  That PR will be merged into this one gradually.


ItsNickBarry:
Still not sure whether inheritance order is correct, but it's now consistent across all contracts and works for all of them except for `SolidStateERC4626`.  In some cases, `ERC165` interfaces had to be removed.

ItsNickBarry:
Tried reversing inheritance order with this task but didn't work:

```ts
import fs from 'fs';
import { task } from 'hardhat/config';

task('reverse').setAction(async (args, hre) => {
  const names = await hre.artifacts.getAllFullyQualifiedNames();

  for (const name of names) {
    const { sourceName } = await hre.artifacts.readArtifact(name);

    const file = fs.readFileSync(sourceName).toString();

    const m = file.match(/(contract|interface)\s+(\w+)\s+is(\s+\w+,?)+\s+{/);

    if (m) {
      const mm = m[0].replaceAll('\n', ' ').match(/\s+is\s+(.*)\s+{/m);
      if (!mm) {
        console.log('=======================');
        console.log(m[0]);
        console.log(m[0].replaceAll('\n', ' '));
      }
      const array = mm![1].trim().split(/\s+/);

      if (array.length > 1) {
        console.log(sourceName);
        array.reverse();

        const repl =
          m[1] +
          ' ' +
          m[2] +
          ' is ' +
          array.map((el) => el.split(',')[0]).join(', ') +
          ' {';
        const fullRepl = file.replace(m[0], repl);

        // console.log(fullRepl);
        fs.writeFileSync(sourceName, fullRepl);
      }
    }
  }
});
```

PR #254, opened 2025-03-01
ERC2981 updates

ItsNickBarry:
Needed to rename `ERC2981` to `NFTRoyalty` to disambiguate implementation contract from standard interface.  Added some internal functions and tests.  Storage slot has changed.


ItsNickBarry:
Should wait for #236 so that storage slot is changed only once.

PR #256, opened 2025-03-02
Use Viem

### LINKED ISSUE #231: `TypeChain` is becoming deprecated
As per https://github.com/dethcrypto/TypeChain/commit/4c9773403a2aa66bf6f5fa066e8ef1b00769c71b it is prudent to consider alternatives to `TypeChain` for testing the `SolidState` contract suite.

Perhaps even considering a switch to Foundry, which would deprecate the need for a `TypeChain` alternative.

ItsNickBarry:
closes #231 

required for #92, which is required for #236 

Edit: a workaround has been implemented for #92, so this PR is not immediately required.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,7 +3,6 @@ node_modules/
 yarn-error.log
 lerna-debug.log
 
-typechain-types/
 dist/
 
 artifacts/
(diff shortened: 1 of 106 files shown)


ItsNickBarry:
Apparently the Viem type system doesn't account for contract inheritance.

ItsNickBarry:
Apparently there's no `hardhat-chai-machers` plugin for Viem.

ItsNickBarry:
Apparently there's a `hardhat-typechain` plugin in Hardhat V3, so this might be unnecessary.

PR #260, opened 2025-03-07
Remove `Mock` contracts

ItsNickBarry:
The `Mock` contracts can be replaced by `hardhat-exposed`, cleaning up the contracts directory a bit.

However, two issues:
* Some test files become extremely verbose (`DoublyLinkedList.ts`).  This could be mitigated by fixing the incompatibility with Typechain and using the default prefix (`$`).
* Some mocks contain special test functions (`ReentrancyGuardMock.sol`) which cannot be reproduced by `hardhat-exposed`.


ItsNickBarry:
Merging into `hardhat-exposed` because I'm now sure that I want to move forward with this approach.  Remaining mocks will be removed from that branch.

PR #261, opened 2025-03-07
Move all implementation to internal contracts

ItsNickBarry:
This PR finally removes all `public` and `external virtual` function declarations, and moves all implementation code to internal contracts


ItsNickBarry:
There might still be some `internal` functions which are not marked as `virtual`.

PR #263, opened 2025-03-08
Allow no unimplemented functions

ItsNickBarry:
Contracts with unimplemented functions cannot be tested via `hardhat-exposed`, so this PR adds missing implementations.

Adding implementation accessors to `Proxy` made the existing `UpgradeableProxy` redundant, so `UpgradeableProxyOwnable` has been merged into it.

`ERC165` functions are still missing from some contracts and need to be added before merge.


ItsNickBarry:
`ERC165` added to `ERC2981` (`NFTRoyalty`) on separate branch in commit fcd64bd6629fced89536505baf0740f1ab2d0cc9.

PR #283, opened 2025-03-17
Array utils and pagination

### LINKED ISSUE #152: Storage lookup pagination
Most of our data structure libraries are represented by arrays internally, and some have a `toArray` function which returns the underlying data.  We need to add a paginated version of `toArray` which accepts `startIndex` and `length` parameters and returns a slice of the underlying array.  This will enable partial lookups when a full lookup would be gas-prohibitive.

Apply this to all current and pending data structures (see `data-structures` branch).  Also apply to contracts that use these data structures, such as the `Enumerable` extensions.

### LINKED ISSUE #163: Add array casting functions
The code in #161 could be encapsulated in array library functions for general use.

### LINKED ISSUE #181: Add `toArray` function to `DoublyLinkedList`
Include pagination: #152.

Paginated version will need to accept an initial value as an input rather than an index, because linked list values are not natively indexed.

ItsNickBarry:
closes #152 
closes #163 
closes #181


ItsNickBarry:
The `PackedDoublyLinkedList` functions could share an internal implementation if it used a `bytes32[]`:
```solidity
function _toArray(
    _PackedDoublyLinkedList storage self,
    bytes16 prevValue,
    uint256 count
) internal view returns (bytes32[] memory array) {
    array = new bytes32[](count);

    for (uint i; i < count; i++) {
        (, bytes16 nextValue) = _parseLinks(self._links[prevValue]);

        if (nextValue == 0) {
            if (i == 0 && _prev(self, 0) != prevValue)
                revert PackedDoublyLinkedList__NonExistentEntry();

            // truncate the array if end of list is reached
            assembly {
                mstore(array, i)
            }

            break;
        }

        prevValue = nextValue;

        assembly {
            mstore(
                add(32, add(array, mul(i, 32))),
                or(nextValue, shr(128, nextValue))
            )
        }
    }
}
```

But I don't know whether this could cause problems or not.

PR #284, opened 2025-03-17
Add `ERC721` and `ERC1155` extension interfaces

ItsNickBarry:
The `Metadata` and `Enumerable` extensions are defined in the EIPs, so their interfaces should be included in `interfaces/`.


ItsNickBarry:
Also updated URI generation to better match the EIP spec.

PR #286, opened 2025-03-17
Use custom errors in `AddressUtils`

ItsNickBarry:
This significantly reduces the size of contracts that use the `AddressUtils` functions.  In particular, `Proxy` in #226.

### DIFF
--- a/contracts/utils/AddressUtils.sol
+++ b/contracts/utils/AddressUtils.sol
@@ -13,12 +13,6 @@ library AddressUtils {
     error AddressUtils__FailedCall();
     error AddressUtils__FailedCallWithValue();
 
-    // TODO: without this function, custom errors referenced by selector are not available in tests
-    function __placeholder() internal {
-        revert AddressUtils__FailedCall();
-        revert AddressUtils__FailedCallWithValue();
-    }
-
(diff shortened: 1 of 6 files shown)


contracts/utils/AddressUtils.sol
```diff
@@ -10,6 +10,14 @@ library AddressUtils {
...
     error AddressUtils__NotContract();
     error AddressUtils__SendValueFailed();
+    error AddressUtils__FailedCall();
+    error AddressUtils__FailedCallWithValue();
+
+    // TODO: without this function, custom errors referenced by selector are not available in tests
+    function __placeholder() internal {
+        revert AddressUtils__FailedCall();
+        revert AddressUtils__FailedCallWithValue();
+    }
```
ItsNickBarry, line 20: This may be a problem with `hardhat-chai-matchers`, as the errors appear in the `AddressUtils` ABI even without this function.

ItsNickBarry: Actually, they don't appear in `$AddressUtils`, which is the contract used in the tests.  They're not inherited because `AddressUtils` is a `library`.  I think this is a Solidity issue, though arguably it's expected behavior.

PR #287, opened 2025-03-18
Add `receive` function to `Proxy`

ItsNickBarry:
This change suppresses several compiler warnings.  Need to check `DiamondProxy` tests before merging.


ItsNickBarry:
The tests are passing, but the spec interface isn't flexible enough to test all situations.

PR #292, opened 2025-03-22
Conform proxies to `ERC1967` and add transparent proxies

ItsNickBarry:
Proxies no longer use `Ownable` for upgrades (includeing `diamondCut`).

Storage layout seed string format has been changed to `solidstate.layout.[name]` because I thought it would be funny if `ERC1967Storage` complied with both `ERC1967` and `ERC7201`, and I needed the resulting value to be less than the `ERC1967` `implementation` slot.


ItsNickBarry:
I would like to add a UUPS-style proxy, but [ERC-1822](https://eips.ethereum.org/EIPS/eip-1822) is not a real standard, and it contradicts ERC-1967.

I would also like to move the transparent proxy `admin` functions into a shared module, but there's no logical place to put it.

ItsNickBarry:
It would be nice if `DiamondBeacon` inherited from `Beacon`, but that would involve appending function signature as metadata to the `implementation()` function.  Might make sense if a beacon equivalent of `DiamondFallback` were added.

PR #294, opened 2025-03-26
Add `Slot` data types

ItsNickBarry:
This also adds safe casting functions to the primitive type utility libraries, which are necessary for interacting with `Slot` types.

Probably want to move `calculateErc7201StorageSlot` to `Slot.sol` as well, but the function isn't even use anywhere (waiting for https://github.com/ethereum/solidity/issues/3157).


ItsNickBarry:
Renamed utility libraries to make interacting with slots cleaner.

ItsNickBarry:
Need to include:
* `offset` parameter for read/write functions for dealing with packed data
* `adjacent` or similar function to get nearby slots

If typed write functions were reintroduced, multi-type functions could also be included for tight packing.  For example:

```solidity
function write(StorageSlot slot, address _0, bool _1) internal;
```

ItsNickBarry:
Multi-type functions are not possible due to the number of types.  Alternative added in #295.  That PR also handles slot offsets, in a way.

Casting functions will need to be improved later, probably.

PR #295, opened 2025-03-27
`Bytes32Builder`

ItsNickBarry:
Like Java `StringBuilder`, but worse.

Alternative to `abi.encodePacked` for packing small primitive types into storage slots.  Differences:
* reversed insertion order to better match storage layout
* unpacking functions
* slightly cheaper (according to some previous tests)


ItsNickBarry:
Add `splice`/`unsplice`.

PR #303, opened 2025-03-30
Clean up directory structure

ItsNickBarry:
Merge directories and files, fix imports.


ItsNickBarry:
Hardhat 3 might support packages with direct Solidity imports.  If so, we'll update all internal imports later.

```typescript
for (const fullyQualifiedName of fullyQualifiedNames) {
  const [sourceFile, entityName] = fullyQualifiedName.split(':');

  let source = String(await fs.promises.readFile(sourceFile));

  const matches = source.match(/from \'.*\';/g) ?? [];
  console.log('___________________________', entityName);
  for (const match of matches) {
    const m = match.replace("from '", '').replace("';", '');
    console.log(m);
    console.log(path.dirname(sourceFile));
    const fullPath = path.relative(
      hre.config.paths.root,
      path.resolve(path.dirname(sourceFile), m),
    );
    console.log(fullPath);
    console.log();

    source = source.replace(m, fullPath);
  }
  console.log(source);

  await fs.promises.writeFile(sourceFile, source);
}
```

PR #308, opened 2025-04-06
Make `Slot` types free types, include functions globally

ItsNickBarry:


ItsNickBarry:
Might want to split `Slot` into `StorageSlot` and `TransientSlot` or `SSlot` and `TSlot`.

PR #324, opened 2026-06-13
Safe ERC20 permit

ItsNickBarry:
This PR adds a `safePermit` function to the `SafeERC20` library.  It allows transaction execution to continue even if a permit has been front-run as part of a DoS attack, to which ERC2612 is susceptible.  Includes tests for some other `SafeERC20` functions.


ItsNickBarry:
Also changes `FungibleTokenPermit` inheritance to include all `FungibleToken` (`ERC20`) functions.  The permit contract no longer works standalone.

